PR #109, opened 2018-10-01
104 creators and contributors updates

cdmo:
required for https://git.psu.edu/i-tech/psulib_traject/merge_requests/2 to work


banukutlu:
we decided to remove `marc_indexer.rb` since we will not be using `rake solr:marc:index_test_data` to index marc record but only run traject cmd with the config from the psulib_traject repo to index records.

banukutlu:
https://github.com/psu-libraries/psulib_blacklight/blob/master/app/models/solr_document.rb#L13

PR #112, opened 2018-10-03
88 setup continuous integration

cdmo:
this puts into place Travis CI integration testing using niftany (rubocop, erb and scss) linting. There will be todos after this in terms of cleanup to address violations.

I did run a `bundle exec rubocop -a` which fixed a lot of the low-hanging violations, hence so many files changed.


cdmo:
Not sure why Travis is running twice here. Will check into

cdmo:
I think I'll probably cancel this sucker..

Issue #136, opened 2018-10-17
yarn integrity


cdmo:
from deployment script:

```
TASK [ansible-rails_app : checkout the rails app code from source control] ***********************************
fatal: [blackcat01qa.vmhost.psu.edu]: FAILED! => {"before": "aeb8d4aa1339bb03c4335bd8cd29024234f4b268", "changed": false, "msg": "Local modifications exist in repository (force=no)."}
```

related to https://git.psu.edu/i-tech/psulib_blacklight_deploy/issues/17

cdmo:
this is due to the fact that rails runs an integrity check on production when it does a yarn install as part of deployment, I believe

cdmo:
ok, so when i run `yarn install --integrity` and then `rails s` locally, all is kosher.

when I run `rails webpacker:yarn_install` and then `rails s` locally, I get the following error:

```
➜  psulib_blacklight git:(#136-yarn-integrity) ✗ rails s
=> Booting Puma
=> Rails 5.2.1 application starting in development
=> Run `rails server -h` for more startup options


========================================
  Your Yarn packages are out of date!
  Please run `yarn install` to update.
========================================


To disable this check, please add `config.webpacker.check_yarn_integrity = false`
to your Rails development config file (config/environments/development.rb).


yarn check v1.10.1
warning Integrity check: Flags don't match
error Integrity check failed
error Found 1 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.


Exiting
```

Looking at the rails task for yarn_install it does this:

`yarn install --no-progress --production`

So, it runs that production flag. This means 

> "Yarn will not install any package listed in devDependencies"
https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production-true-false

Which is great for production, obviously, but not so great for dev. It's also tangent to the original problem of this issue: integrity checks. 

After speaking with @banukutlu it looks like the best option might be to just modify `webpacker.yml` a la https://github.com/rails/webpacker#yarn-integrity

cdmo:
I think what's going on here is that the two settings (one in `webpacker.yml` and one in `development.rb` or `production.rb`) control two different things:

1. At the `rb` level, rails is being told to check integrity of the install yarn packages or not.
2. At the yml config `yarn` is being told whether or not to generate the integrity digests or not.

[...]

PR #140, opened 2018-10-17
#136 yarn integrity

cdmo:
let's try this, will instruct rails to run yarn install with --freeze-lockfile flag


banukutlu:
I ran a deployment and looks like yarn.lock has not been modified. It looks like overriding the yarn install task worked.

cdmo:
huzzah!

PR #170, opened 2018-11-05
#165 docker solr

cdmo:
this removes solr_wrapper entirely and replaces it locally with docker solr.

@banukutlu can you give this a spin locally to test it out before merging? you'll need to download [docker ce for mac](https://store.docker.com/editions/community/docker-ce-desktop-mac), including creating an account. Note that the Procfile is updated too. I think that after you download and install, then checkout this branch you should be able to run our foreman command. After that you'll import using traject as usual.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,6 +17,7 @@ Metrics/BlockLength:
   - 'app/controllers/catalog_controller.rb'
   - 'db/**/*'
   - 'spec/**/*'
+  - 'lib/tasks/**/*'
 
 Metrics/LineLength:
   Exclude:
--- a/.solr_wrapper.yml
+++ b/.solr_wrapper.yml
@@ -1,9 +1,0 @@
-# Place any default configuration for solr_wrapper here
-# port: 8983
-version: 7.4.0
-collection:
-  dir: solr/conf/
-  name: blacklight-core
-  persist: true
-instance_dir: tmp/solr/
-download_dir: tmp/solr_dl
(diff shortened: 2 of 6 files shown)


banukutlu:
@cdmo Getting this when I run our foreman cmd. Am I missing anything? I installed docker ce, checked out the branch and run bundle install as well 🤔 

```
10:30:13 solr.1      | Error: No such object: felix
10:30:14 solr.1      | 7.4.0: Pulling from library/solr
10:30:14 solr.1      | Digest: sha256:e99a84c8fb8b457a9f911db3500afeeed010f59ed45f0caac51d79c848950470
10:30:14 solr.1      | Status: Image is up to date for solr:7.4.0
10:30:14 solr.1      | invalid argument "-p" for "-a, --attach" flag: valid streams are STDIN, STDOUT and STDERR
10:30:14 solr.1      | See 'docker run --help'.
10:30:14 solr.1      | CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
10:30:14 solr.1      | exited with code 0
```

cdmo:
it's almost *too* easy...

cdmo:
you might want to try the clean command too `bundle exec rails docker:clean`. Also check if data persists after running `bundle exec rails docker:down`?

Again, the biggest pain to me is that you have to run that docker down command after killing the foreman task.

cdmo:
btw Felix the Cat is in the public domain 

felix-the-cat-image-by-otto-messmer

banukutlu:
data persisted after stop and restart 👍

banukutlu:
> you might want to try the clean command too `bundle exec rails docker:clean`. Also check if data persists after running `bundle exec rails docker:down`?
> 
> Again, the biggest pain to me is that you have to run that docker down command after killing the foreman task.

what happens if we do not or forget to run docker down?

cdmo:
solr just goes on 🤷‍♂️ , you'd have to kill it manually (a restart would kill it too)

banukutlu:
we need to update the docs.

PR #200, opened 2018-12-10
Bound with

cdmo:
tied to https://github.com/psu-libraries/psulib_traject/pull/65

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -143,7 +143,8 @@ class CatalogController < ApplicationController
     config.add_show_field 'isbn_ssim', label: 'ISBN', separator_options: separator_options
     config.add_show_field 'subject_topic_facet_ssim', label: 'Topic'
     config.add_show_field 'id', label: 'Catkey'
-    config.add_show_field 'bound_with_ssm', label: 'Bound in'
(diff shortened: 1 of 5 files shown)


cdmo:
I'll write a test for the helper 😅

banukutlu:
@cdmo just wanted to make sure if this page https://github.com/psu-libraries/psulib_blacklight/wiki/Exporting-and-Displaying-Bound-With-in-Linked-Notes contains your latest work? Binding Notes field to be added right?

cdmo:
I'll update that wiki too if need be. I'll double check.

cdmo:
I added a test and just stubbed the data I needed that the method expects. I'd like to hear your feedback.

cdmo:
I updated the wiki to add some clarity/specificity.

However, we are not indexing this field beyond the `all_text_tsiv` fields as far as I know and this is on purpose. Can I change the wiki to reflect this or do we need to add an `i` to the dynamic field? @banukutlu

cdmo:
Updated the wiki to clarify the indexing of bound with.

cdmo:
I don't get what Coveralls is so mad about, https://coveralls.io/builds/20600061, I didn't even touch anything in `/models` 🤷‍♂️

PR #219, opened 2019-01-30
#46 advanced search

### LINKED ISSUE #46: Advanced Search
Wiki documentation: https://github.com/psu-libraries/psulib_blacklight/wiki/Advanced-Search

https://github.com/projectblacklight/blacklight_advanced_search

### LINKED ISSUE #210: blacklight gem to 7.0.1 from 7.0rc2
https://github.com/projectblacklight/blacklight/releases/tag/v7.0.0

cdmo:
Closes #46 
This is probably ready to go. Will close #210 too. One caveat is that we probably don't need the sass from the blacklight-advanced-search gem itself, but we can revisit that in mvp 3 I'm thinking.


app/controllers/catalog_controller.rb
```diff
@@ -178,41 +191,57 @@ class CatalogController < ApplicationController
...
 
     config.add_search_field('title') do |field|
-      # solr_parameters hash are sent to Solr as ordinary url query params.
       field.solr_parameters = {
-        'spellcheck.dictionary': 'title',
         qf: '${title_qf}',
         pf: '${title_pf}'
       }
+      field.solr_adv_parameters = {
+        qf:  '$qf_title',
```
banukutlu: should this be `$qf_title` or `$title_qf`? 
and same question for pf?

solr/conf/solrconfig.xml
```diff
@@ -143,6 +143,12 @@
          subject_addl_tsim^10
        </str>
 
+       <!-- for advanced search number text box -->
+       <str name="qf_number">
```
banukutlu: If we will keep the naming for others as `*_qf`, what do you think changing this to `number_qf`. Or we can change the others as `qf_*`?

banukutlu: Do we need to add a `pf_number` for boosting?

app/controllers/catalog_controller.rb
```diff
@@ -9,6 +10,18 @@ class CatalogController < ApplicationController
...
 
   configure_blacklight do |config|
+    # default advanced config values
+    config.advanced_search ||= Blacklight::OpenStructWithHashAccess.new
+    # config.advanced_search[:qt] ||= 'advanced'
+    config.advanced_search[:url_key] ||= 'advanced'
+    config.advanced_search[:query_parser] ||= 'edismax'
+    config.advanced_search[:form_solr_parameters] ||= {
+      # 'facet.field' => ['language_facet_ssim'], # will add 'library', 'location', 'material_type'
+      'facet.limit' => -1, # return all facet values
```
banukutlu: Subjects listed on advanced search seems to be listing only 20 subjects because of `limit: 20` here: https://github.com/psu-libraries/psulib_blacklight/blob/%2346-Advanced-Search/app/controllers/catalog_controller.rb#L112. 

`'facet.limit' => -1` supposed to be taking take of this, I guess? @cdmo could you also check pls? I tried adding `"f.subject_topic_facet_ssim.facet.limit" => -1,` to `config.advanced_search[:form_solr_parameters]` and it returned all. 

Similar behavior for author and language dropdowns as well since there is also a limit defined for these facets.

app/controllers/catalog_controller.rb
```diff
@@ -178,41 +191,57 @@ class CatalogController < ApplicationController
...
+      field.label = 'ISBN/ISSN'
+      field.solr_parameters = {
+        qf:  '$qf_number'
       }
     end
 
+    ## Stubs for Advanced Search as per https://github.com/psu-libraries/psulib_blacklight/wiki/Advanced-Search
+    # config.add_search_field('Series')
+    # config.add_search_field('Publisher')
+    # config.add_search_field('Publication date')
```
banukutlu: `Publish Date` search from advanced is failing because `config.add_search_field('Publication date')` needs to be implemented, right? I think we talked about it but I just wanted to confirm. This is to be handled with the publication date changes.

cdmo: Didn't test that one, again, mainly because I am not sure what fields will end up in our list. Here's the error I see when testing that facet:

```
2019-02-01 13:44:35.925 ERROR (qtp1107530534-16) [   x:blacklight-core] o.a.s.h.RequestHandlerBase org.apache.solr.common.SolrException: undefined field pub_date_facet_field
```

cdmo: We changed this to `pub_date_ssim`

cdmo: My vote is to not fix this unless it turns out that we need to support it based on Ruth's input. What do you think?

app/controllers/catalog_controller.rb
cdmo: For now I propose that we just comment out these custom configs. Especially the byte-order one.

We will likely have to do some more customization, like displaying the number of items for a given facet.

cdmo:
Ready to be re-reviewed, I think.

app/controllers/catalog_controller.rb
```diff
@@ -178,41 +191,57 @@ class CatalogController < ApplicationController
...
-        qf: '${subject_qf}',
-        pf: '${subject_pf}'
+        qf: "'${subject_qf}'",
+        pf: "'${subject_pf}'"
+      }
+    end
+
+    config.add_search_field('identifiers') do |field|
+      field.include_in_simple_select = false
+      field.label = 'ISBN/ISSN'
```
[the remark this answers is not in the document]
banukutlu: I agree and I asked this even though I knew how your response will be :) Yes, Stanford has "ISBN, ISSN, etc".

app/controllers/catalog_controller.rb
banukutlu: Yes our field is `pub_date_ssim`. 
I just checked again and publication date is listed on the advanced search fields on wiki. 
Saying that I think it does not have to be fixed with this PR, sth I need to fix with publication fields changes, right? I just wanted to make sure if you are also getting an error and also wanted to confirm that this is to be fixed later.

app/controllers/catalog_controller.rb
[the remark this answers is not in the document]
cdmo: lol

app/controllers/catalog_controller.rb
cdmo: There are just a lot ways we could handle publication date search/facet. Exact date with checkbox? Checkbox range? A true range picker like many other institutions have adopted? Etc.

app/controllers/catalog_controller.rb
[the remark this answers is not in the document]
banukutlu: so we are to leave it as is for QA review then?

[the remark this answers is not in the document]
banukutlu: yeah, I think we should let Ruth play around and then give us feedback for further customizations. (ps. I know 😄)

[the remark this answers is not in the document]
cdmo: Exactly, yes

solr/conf/solrconfig.xml
cdmo: The short answer is "I don't know." I did think about it though and my thoughts were that we are only looking at two fields for this search. Two numbers-only fields. So what good would a phrase field be? I would assume there aren't even any spaces in these fields. ?

Issue #204, opened 2018-12-12
Index and display publication fields


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Publication-Info-Fields-Index-and-Display

Requires  psu-libraries/psulib_traject#18
Requires psu-libraries/psulib_traject#31

banukutlu:
@ruthtillman, we have a `Publication Year` facet but are we gonna keep the `Publish Date` facet?
![Screen Shot 2019-02-01 at 11.45.03 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5bae8d163be62e0012255ac7/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5700b77216454f26b7191a379ee0c2f54415f4dbc30254d58eeb9cb74fce36f34985619e8833c0a7c923e19afb98719bb57ba2083468cbfeadf44c8fa2137a5dad11b61435c52dcbd4f3fb29842c259f34fe84d56ad163064e173306e0a.png)

ruthtillman:
No, we won't.

ruthtillman:
While it technically works very well, the number of years makes it unwieldy as currently laid out. Can we implement the range boxes similar to what Stanford and Princeton have. IDEALLY without the slider.

If we have to do the slider, let's start date at 1800

banukutlu:
> While it technically works very well, the number of years makes it unwieldy as currently laid out. Can we implement the range boxes similar to what Stanford and Princeton have. IDEALLY without the slider.
> 
> If we have to do the slider, let's start date at 1800

@ruthtillman I  added #234 for implementing the range on pub year, so I guess we can close this issue?

PR #221, opened 2019-02-06
#204 publication fields

banukutlu:
Connected to #204 

- index and display publication fields
- to hide the publication statement label in search results added `psul_index_presenter.rb` and `_index_default.html.erb` partial (there might be a better way to solve this)
- also reordered the single item labels

### DIFF
--- a/spec/presenters/psul_index_presenter_spec.rb
+++ b/spec/presenters/psul_index_presenter_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PsulIndexPresenter do
+  subject(:presenter) { described_class.new(document, request_context, config) }
+
+  let(:document) { SolrDocument.new }
+  let(:request_context) { instance_double(ActionView::Base, document_index_view_type: 'list') }
+  let(:config) { Blacklight::Configuration.new }
+
+  describe '#label' do
+    context 'when a document has a publication statement' do
+      let(:document) do
+        SolrDocument.new(id: 1,
+                         format: 'Book',
+                         publication_display_ssm: 'This is the publication statement of the document')
+      end
+
+      let(:config) do
+        Blacklight::Configuration.new.configure do |config|
+          config.add_index_field 'publication_display_ssm'
+          config.add_index_field 'format'
+        end
+      end
+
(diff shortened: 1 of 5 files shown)


spec/presenters/psul_index_presenter_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+
+      let(:config) do
+        Blacklight::Configuration.new.configure do |config|
+          config.add_index_field 'publication_display_ssm'
+          config.add_index_field 'format'
+        end
+      end
+
+      it 'publication statement has an empty label' do
```
cdmo: Maybe this should read like `it 'should not have a label for publication statement, but still have a label for format' do` ?

banukutlu:
> Maybe just a small change to the spec's language. Other then that it looks great. I can't think of a more elegant way to handle this honestly. And it might become a useful avenue for future front-end work and tweaks.
> 
> I feel like to do this "right" we'd have to do a significant amount of work in re-working BL itself. Like, a new conf setting could be added for "show_label" with a value of "true|false|record|results" or something like that and then add a method to create the html itself (i.e., `content_tag()`) based on that toggle.

💯

spec/presenters/psul_index_presenter_spec.rb
banukutlu: sure! I just changed it but had to leave `should` out because rubocop was not happy with it.

cdmo:
wonder why you merged in master too?

banukutlu:
> wonder why you merged in master too?

there were conflicts, I used GH's Resolve Conflicts then commit

spec/presenters/psul_index_presenter_spec.rb
cdmo: will have to think about that one!

cdmo:
> > wonder why you merged in master too?
> 
> there were conflicts, I used GH's Resolve Conflicts then commit

Ah ok, interesting.

I've been loving rebase lately. But, rebasing here would mean a new PR.

spec/presenters/psul_index_presenter_spec.rb
banukutlu: 😄

Issue #218, opened 2019-01-28
Index and display languages field


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Language-Fields-Index-and-Display

Requires psu-libraries/psulib_traject#79

banukutlu:
@ruthtillman are we gonna add `041` fields to language facet and display?

ruthtillman:
You're correct, I updated it.

Issue #202, opened 2018-12-12
Index and display note fields


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Note-Fields-Index-and-Display

Requires https://github.com/psu-libraries/psulib_traject/pull/15

banukutlu:
@ruthtillman the order of notes fields here:https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields and https://github.com/psu-libraries/psulib_blacklight/wiki/Note-Fields-Index-and-Display are a bit different. Which order is the we should use?

ruthtillman:
For ordering fields generally, let's stick with Blacklight-Catalog-Fields.

cdmo:
@ruthtillman [Note-Fields-Index-and-Display](https://github.com/psu-libraries/psulib_blacklight/wiki/Note-Fields-Index-and-Display) shows "Location of Originals" and [Blacklight-Catalog-Fields](https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields) shows "Location of Original"

cdmo:
in the same vain of questions, should it be "Associated" or "Associated Materials"?

cdmo:
Administrative Note, or Administrative History?

cdmo:
Is "Reviewed/Cited In" the same as "References"?
Is "Table of Contents" the same as "Contents"?
Is "Summary" the same as "Abstract"?

And these don't appear to be in the Blacklight-Catalog-Fields page at all:

"Scale Note for Graphic Material"
"Indexed By"
"Selectively Indexed By"
"References"
"Summary" (unless this is "Abstract")
"Reading Grade"
"Interest Age"
"Interest Grade"
"Related Materials"
"Language Note"  

Finally, @banukutlu related, these are in our system, but aren't in the ordering page. What should we do with them?

"Catkey"
"Call number"
"URL"
"More Information"

banukutlu:
@cdmo we will remove Catkey (I think we should just remove it now - it is already in the url if needed for testing purposes). Call number should be removed as well. 
@ruthtillman what about URL and More information?

ruthtillman:
@banukutlu  ++ on removing catkey (we'll put it... somewhere on the page but not in the main area) I'll get on URL when I get back to workplace.

@cdmo: I think I've properly standardized these across the main fields + the note wiki index... and then ensured they were in the single record field list.

[...]

PR #224, opened 2019-02-08
#202 note fields

cdmo:
Connected to #202 
Connected to #218

Related to https://github.com/psu-libraries/psulib_traject/pull/95


app/controllers/catalog_controller.rb
```diff
@@ -167,6 +167,58 @@ class CatalogController < ApplicationController
     config.add_show_field 'url_suppl_display_ssm', label: 'More Information'
     config.add_show_field 'lc_callnum_display_ssm', label: 'Call number'
     config.add_show_field 'id', label: 'Catkey'
+    config.add_show_field 'general_note_ssm', label: 'Note'
```
banukutlu: @cdmo it is a minor thing but would you mind changing the order of the fields depending on here: https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields#single-item-view-labels?

Issue #220, opened 2019-02-05
Index and display formats


ruthtillman:
https://github.com/psu-libraries/psulib_blacklight/wiki/Formats

These will display as a facet in the sidebar and be used to display some info about the record, e.g. map to an item type.

banukutlu:
@ruthtillman do we want a label on search results for format?

ruthtillman:
In the long run, I'd like to do an icon or such. In the short run, a label "Format" -- sure.

banukutlu:
@ruthtillman I added fontawesome so won't be displaying a label on search results. What about single item view? I guess we will have the label there - is it gonna be `Format`? Also are we gonna display an icon on single item view?

ruthtillman:
@banukutlu can we use the similar layout to Princeton? So -- icon + format name (or occasionally names?) on search result. Then icon as thumbnail (until we get real good thumbnails) -- and Format: Book or whatever... on the single page.

banukutlu:
@ruthtillman any icon for default format "Other"?

ruthtillman:
@banukutlu no real preference for other. Possibly: https://fontawesome.com/icons/scroll or https://fontawesome.com/icons/file ? ... I was tempted to suggest https://fontawesome.com/icons/otter for the sound-alike, but...

ruthtillman:
Updated Games logic to cut out some possible mistakes when the record 008 is bad but the main leader value is still ok.

banukutlu:
> Updated Games logic in formats wiki page to cut out some possible mistakes when the record 008 is bad but the main leader value is still ok.

This will be handled with psu-libraries/psulib_traject#113

ruthtillman:
Closing this issue. We may revisit with specific parameters as we go forward, but first pass of formats is done.

PR #238, opened 2019-03-11
#220 format fields

banukutlu:
Connected to #220


app/views/catalog/_show_sidebar.html.erb
```diff
@@ -0,0 +1,16 @@
+<% unless @document[:format][0].empty? %>
```
cdmo: Does every record have a format? If it didn't would this it throw an error? Is every format an array? If it wasn't would it cause an error because [0] doesn't exist? I ask just in case you've already checked.

app/views/layouts/blacklight/base.html.erb
```diff
@@ -32,6 +32,17 @@
 
   <%= render partial: 'shared/flash_msg', layout: 'shared/flash_messages' %>
 
+  <% if @search_context %>
```
cdmo: So, I'm curious as to why you had to move the search context bit and the previous next doc bit here. Seems unrelated to the format work, but I'm missing something, right?

app/helpers/application_helper.rb
```diff
@@ -66,4 +66,9 @@ def genre_links(options = {})
   def display_duration(options = {})
     options[:value]&.map { |v| v.scan(/([0-9]{2})/).join(':') }
   end
+
+  # To render format icon on search results as the default thumbnail for now
+  def render_thumbnail(document, _options = {})
```
cdmo: Curious about why you're using document as opposed to options (i.e., `options[:value]`)

app/views/catalog/_show_sidebar.html.erb
banukutlu: yes, every record should have a format, if none of the rules turns a format then it falls to default "Other". And it should always be an array, Of course both of these are expected unless there is a bug in the logic :)

app/helpers/application_helper.rb
banukutlu: I could not get it working with options[:value] :(

app/views/layouts/blacklight/base.html.erb
banukutlu: after adding thumbnail, pagination widgets and start over did not look good in the content section so I moved them out here so they expand over both sections.

app/helpers/application_helper.rb
cdmo: I guess because of the scope at which this is called. Okay, learned something new! :)

app/views/layouts/blacklight/base.html.erb
cdmo: that makes sense! nice work, thanks.

Issue #225, opened 2019-02-14
Index and display media types


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Media-Types

ruthtillman:
Reviewed, good!

PR #254, opened 2019-04-01
#225 media types

banukutlu:
Connected to #225


.rubocop_todo.yml
```diff
@@ -23,7 +23,7 @@ RSpec/ExpectActual:
     - 'spec/routing/**/*'
     - 'spec/controllers/catalog_controller_spec.rb'
 
-# Offense count: 2
+# Offense count: 3
```
cdmo: I don't see any new specs, so how could their be a new violation?

banukutlu: probably it passed with this exception but todo file not updated

cdmo: I didn't think that was possible. Like, it should have failed rubocop's inspection before a merge 🤷‍♂️

Issue #248, opened 2019-03-21
Enable docValues for faceting and sorting fields, and general cleanup/tune of facets


banukutlu:
- [x] Enable docValues for all for faceting and sorting field. Currently not all string fields have it enabled.
- [x]  Remove stored=true from fields that have docValues enabled (including string and numeric fields) and use docValues to retrieve the original data. To do that, either move the schema to 1.6 or to add useDocValuesAsStored for each field type. 

Some refactoring based on audit of facets:
- [x] genre facet can be sim
- [x] subject topic facet can be sim
- [x] remove  "all_authors_facet_ssim"
- [x] remove "genre_full_facet_ssim" 

- [x] review access whence merged

cdmo:
this has been merged in and there's not much to QA because it's all under the hood stuff. So closing.

PR #255, opened 2019-04-02
#248 docvalues

cdmo:
This is really a bit of an experiment, so we can excise these commits if the test shakes out poorly. Locally I'm not seeing any difference in terms of display and search. It is snappy, but that might be because of a small sample size.


solr/conf/solrconfig.xml
```diff
@@ -170,19 +170,18 @@
...
          format,
          publication_display_ssm,
-         pub_date_ssim,
+         pub_date_ssm,
          overall_imprint_display_ssm,
          copyright_display_ssm,
          edition_display_ssm,
          phys_desc_ssm,
-         language_facet_ssim,
+         language_ssim,
```
banukutlu: @cdmo should this be `language_facet`? 🤔

app/controllers/catalog_controller.rb
```diff
@@ -133,24 +131,24 @@ class CatalogController < ApplicationController
...
+    config.add_show_field 'author_meeting_display_ssm', label: 'Conference Author', link_to_facet: :all_authors_facet
     config.add_show_field 'uniform_title_display_ssm', label: 'Uniform Title'
     config.add_show_field 'additional_title_display_ssm', label: 'Additional Titles'
     config.add_show_field 'format', label: 'Format'
     config.add_show_field 'overall_imprint_display_ssm', label: 'Published'
     config.add_show_field 'copyright_display_ssm', label: 'Copyright Date'
     config.add_show_field 'edition_display_ssm', label: 'Edition'
     config.add_show_field 'phys_desc_ssm', label: 'Physical Description'
-    config.add_show_field 'language_facet_ssim', label: 'Language'
+    config.add_show_field 'language_ssim', label: 'Language'
```
banukutlu: @cdmo should this be `language_facet`?

cdmo: No because we display language in records and we also include it in our index as part of search, so, I split the field into two: one for faceting and one for indexing/displaying. See catalog controller.

solr/conf/schema.xml
```diff
@@ -194,14 +202,30 @@
...
+    <field name="format"              type="string" docValues="true" stored="false" multiValued="true"/>
+    <field name="genre_facet"         type="string" docValues="true" stored="false" multiValued="true" />
+    <field name="genre_full_facet"    type="string" docValues="true" stored="false" multiValued="true" />
+    <field name="language_facet"      type="string" docValues="true" stored="false" multiValued="true" />
+    <field name="lc_1letter_facet"    type="string" docValues="true" stored="false" multiValued="true" />
+    <field name="pub_date_facet"      type="string" docValues="true" stored="false" multiValued="true" />
+    <field name="subject_topic_facet" type="string" docValues="true" stored="false" multiValued="true" />
+
+    <!-- Sorts -->
+    <field name="pub_date_sort"       type="string" docValues="true" stored="false" multiValued="true" />
```
banukutlu: @cdmo I think `pub_date_sort_itsi` had  docValues enabled already.  [`type: tint`](https://github.com/psu-libraries/psulib_blacklight/blob/8db6bcc9d5be9cc8c3a5c9b581dfb3b79c4bd7db/solr/conf/schema.xml#L98) is defined as  [`<fieldType name="tint" class="solr.IntPointField" docValues="true"/>`](https://github.com/psu-libraries/psulib_blacklight/blob/8db6bcc9d5be9cc8c3a5c9b581dfb3b79c4bd7db/solr/conf/schema.xml#L287).

solr/conf/solrconfig.xml
cdmo: the facet is now docvalues. For the sake of just testing docvalues, I split language into two fields, one for faceting and one for index and display. Same for pub date

app/controllers/catalog_controller.rb
banukutlu: I see 👍  thanks!

solr/conf/schema.xml
cdmo: Yup, that's right, good catch thanks!

solr/conf/schema.xml
```diff
@@ -194,14 +202,29 @@
...
+    <field name="format"              type="string" docValues="true" indexed="false" stored="false" multiValued="true"/>
+    <field name="genre_facet"         type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
+    <field name="genre_full_facet"    type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
+    <field name="language_facet"      type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
+    <field name="lc_1letter_facet"    type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
+    <field name="pub_date_facet"      type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
+    <field name="subject_topic_facet" type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
+
+    <!-- Sorts -->
+    <field name="title_sort"          type="string" docValues="true" indexed="false" stored="false" multiValued="true" />
```
banukutlu: @cdmo solr doc says for sorting on a field, multivalued should be false: https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html#CommonQueryParameters-ThesortParameter 🤔

cdmo: > If the field is multi-valued, Lucene will use the SORTED_SET type.

from [solr docs](https://lucene.apache.org/solr/guide/7_4/docvalues.html#enabling-docvalues)

So, it appears to be legal - but how it works in practice... not sure.

banukutlu: cool! thanks for the info, I just did not know :)

cdmo:
I had a conversation with some folks on the Blacklight channel in Code4Lib Slack, and it appears that removing the local parameters that add the TermQuery Parser could cause issues. Specifically "then you have to worry about escaping  and a bunch of other complexity that the term parser just handles." Not sure why TermQuery added via local params causes docvalue field facets to return no results. That's where we're at right now.

Locally I have 2M records indexed and I am able to click through on facets items from all of our facets, including ones with special characters and diacritics, for what it's worth. 

I'll mark this PR as WIP for now though.

cdmo:
Ok, got more clarity on this now. We definitely want to keep `indexed=true` for those facets. Otherwise, you're doing a table scan of the column-oriented values for every facet query. Some folks on the BL C4L Slack were very helpful on this topic. They found these two emails from the solr user list:

https://mail-archives.apache.org/mod_mbox/lucene-solr-user/201811.mbox/%3cCAN4YXvcpF+ys2ERwAsOX7VmhfLNrzQNEf3udJxT0KkbcoBSOPQ@mail.gmail.com%3e

https://mail-archives.apache.org/mod_mbox/lucene-solr-user/201811.mbox/%3c2e70e51f-0abc-bd97-c4ba-3c6f1291c9b4@elyograg.org%3e

I'll copy the first email here just for posterity in case the links above die some day:

> I've noticed this  too, but I think it's more a side effect than
> something usable for the reasons you outlined. Searching a docValues
> field is akin to a "table scan", the uninverted structure is totally
> unsuited for searching. It works, but as you've found out it's
> unusably  slow for any decent sized corpus. Whether it should throw
> some kind of error or not is worth discussing.
> 
> Erick
> On Mon, Nov 19, 2018 at 7:20 PM Rahul Goswami <rahul196452@gmail.com> wrote:
> >
> > I am using SolrCloud 7.2.1. My understanding is that setting docvalues=true
> > would optimize faceting, grouping and sorting; but for a field to be
> > searchable it needs to be indexed=true. However I was dumbfounded today
> > when I executed a successful search on a field with below configuration:
> > <field name="myField" type="pint" indexed="false" stored="false"
> > docValues="true"/>
> > However the searches don't always complete and often time out.
> >
> > My question is...
> > Is searching on docValues=true and indexed=false fields supported? If yes,
> > in which cases?
> > What are the pitfalls (as I see that searches, although sometimes
> > successful are atrociously slow and quite often time out)?

Changes coming soon for this PR

solr/conf/schema.xml
cdmo: I don't know either Banu! We're working through it :).

Just to document our talk, we're going to leave this as is for the test run and then do a follow up test run where we change title sorting back to the way it was. 
Thanks!

Issue #239, opened 2019-03-12
Advanced Search updates


banukutlu:
Related to #46

Per https://github.com/psu-libraries/psulib_blacklight/wiki/Advanced-Search:

- Add as text field:
  - [x] Series
  - [x] Publisher
- Change to text field:
  - [x] Subject
  - [x] Publication date (perhaps two boxes?)
- Drop-downs should display every value which exists as a facet:
  - [x] Language
  - [x] Format 
- Add as dropdown
  - [x] Media Type 
  - [x] Library (not indexed yet)
  - ~[ ] Location (not indexed yet)~ see #269
- Remove from drop-down:
  - [x] Genre
  - [x] Call Number
  - [x] Subject (move up to text fields section)
- [x] Selects not populating on first load (turbolink load)
- [x] Change "more options" to "advanced search"

ruthtillman:
Everything is looking great but Call Number doesn't seem to work when selected (either alone or with other fields). (however just the highest-letter call number is sufficient once it works, we don't need to break it down any more)

cdmo:
I have a fix for this that we'll deploy soon, but, just wanted to mention that the reason that the classification facet shows up at all is kind of a fluke/bug. Blacklight Advanced Search automatically includes all pivot facets, I think it's a bug in the relationship between BLAS and BL. Anyway, I mention this because including this option wasn't in the [original spec](https://github.com/psu-libraries/psulib_blacklight/wiki/Advanced-Search) and was actually supposed to be removed according to the punch list at the top of this issue. So we definitely want it right, as just the first letter option. ?

PR #261, opened 2019-04-12
#1 library and location

cdmo:
This has a bit of overlap with https://github.com/psu-libraries/psulib_blacklight/issues/239

Note that I didn't add any specs. I didn't think it was necessary given that there isn't any new logic, except for what was added in the partial. I do wonder if it needs an integration test. What do you think?


cdmo:
@banukutlu I think this is set now -- I talked with @ruthtillman and she suggested that we use the Blacklight Single config option for now for these facets and see how users respond, which I think is a great way forward for now too.

banukutlu:
> This has a bit of overlap with #239
> 
> Note that I didn't add any specs. I didn't think it was necessary given that there isn't any new logic, except for what was added in the partial. I do wonder if it needs an integration test. What do you think?

I think we do not have to add an integration test with this PR yet but as we discussed before we will add them eventually, so let's get this merged. When we are done with the marc mapping, let's revisit this and other customizations we added and add tests? what do you think?

cdmo:
Yeah I think this is okay. I feel like I don't have a good sense yet of how much integration is really worth testing given that the stuff that supplies values to partials is tested. Like I want to avoid just testing for typos.

PR #270, opened 2019-04-18
239 advanced search updates

cdmo:
Will also close #234 

No tests for this code, but again, mostly config stuff. May be some good candidates for integration testing later?

PR #272, opened 2019-04-19
239 advanced search updates

cdmo:
cleared conflicts from #270


app/controllers/search_history_controller.rb
```diff
@@ -3,5 +3,9 @@
 class SearchHistoryController < ApplicationController
   include Blacklight::SearchHistory
 
+  # TODO: Consider deleting the following three lines altogether. The following lines were generated during an install.
+  # Removing these lines has zero effect on the display of constraints.
```
banukutlu: thanks for noting this 👍

solr/conf/schema.xml
```diff
@@ -270,8 +270,6 @@
   <copyField source="author_tsim" dest="author_spell"/>
   <copyField source="subject_tsim" dest="subject_spell"/>
   <copyField source="title_tsim" dest="title_spell"/>
-  <copyField source="pub_date_facet" dest="pub_date_ssm"/>
```
banukutlu: yay for removing unnecessary fields :)

banukutlu:
I am getting this error on my local when I add a start or end date for publication year facet with branch 239 checked out.  try this [url](http://localhost:3000/?utf8=%E2%9C%93&range%5Bpub_date_sort_itsi%5D%5Bbegin%5D=&range%5Bpub_date_sort_itsi%5D%5Bend%5D=&search_field=all_fields&q=&range%5Bpub_date_sort_itsi%5D%5Bbegin%5D=1755&range%5Bpub_date_sort_itsi%5D%5Bend%5D=&commit=Limit). Maybe sth is up with my local, wondering if you are getting this as well.

Screen Shot 2019-04-19 at 10 22 39 PM

cdmo:
That error happened because I made a mistake when manually correcting erb-lint violations.

Also, I think I prefer rebasing a feature being PR'd on top of `master` rather than merging `master` into a feature branch during a PR. That means that a PR has to be closed and reopened, but, I think it's cleaner. What do you think?

banukutlu:
> That error happened because I made a mistake when manually correcting erb-lint violations.
> 
> Also, I think I prefer rebasing a feature being PR'd on top of `master` rather than merging `master` into a feature branch during a PR. That means that a PR has to be closed and reopened, but, I think it's cleaner. What do you think?

agreed. rebasing is cleaner 👍

Issue #268, opened 2019-04-18
Call Numbers indexing and display


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Call-Number-Index-and-Display

PR #273, opened 2019-04-23
#268 call numbers indexing and display

banukutlu:
Connected to #268 
Requires https://github.com/psu-libraries/psulib_traject/pull/131


banukutlu:
> Needs specs

Added #274 for specs.

PR #277, opened 2019-04-24
Removing spellchecker

cdmo:


cdmo:
yeah yay, I 'spose. But hopefully one day we can re-implement and it will work well. It is so handy.

PR #293, opened 2019-05-08
267 solr cloud

cdmo:
To test this locally, you'll need to also checkout the branch being PR'd at https://github.com/psu-libraries/psulib_traject/pull/144

Then:

1. kill felix
2. run the Procfile.dev
3. run `bundle exec rails docker:build` (uploads the config to the embedded zk and creates a collection referencing that)

I had to add a manual extra step for config and collection because `docker run` never finishes execution. So, any tasks that happen after that one don't get called. System calls are blocking.

After this then you can just go back to running that procfile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,9 @@ env:
   - RAILS_ENV=test
 
 before_install:
-- docker pull solr:7.4.0
-- docker run -d -p 8983:8983 -v "$(pwd)"/solr/conf:/myconfig solr:7.4.0 solr-create -c blacklight-core -d /myconfig
+- bundle exec rails docker:pull
+- bundle exec rails docker:run
+- bundle exec rails docker:build
 - docker ps -a
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ before_script:
 - bundle exec rails docker:pull
 - bundle exec rails docker:run_ci
 - bundle exec rails docker:build
-- docker ps -a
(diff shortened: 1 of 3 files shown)


cdmo:
I didn't bother figuring out how to make the collection name a global constant...

config/blacklight.yml
```diff
@@ -12,10 +12,10 @@
 
 development:
   adapter: solr
-  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/blacklight-core" %>
+  url: <%= ENV['SOLR_URL'] || "http://127.0.0.1:8983/solr/psul_blacklight" %>
```
banukutlu: I like psul_blacklight 👌

banukutlu:
worked on my local on the first trial 👍

.travis.yml
```diff
@@ -13,12 +13,11 @@ env:
...
-before_install:
-- docker pull solr:7.4.0
-- docker run -d -p 8983:8983 -v "$(pwd)"/solr/conf:/myconfig solr:7.4.0 solr-create -c blacklight-core -d /myconfig
-- docker ps -a
-
 before_script:
+- bundle exec rails docker:pull
+- bundle exec rails docker:run_ci
+- bundle exec rails docker:build
+- docker ps -a
```
banukutlu: @cdmo do we need this? it just displays the running container, right?

cdmo: it has been there all along... not sure why it's showing it as added. But, anyway, it's not really needed no :)

PR #314, opened 2019-05-20
Refactor template literals

cdmo:
Hopefully the commit messages cover everything. I did also change variable names, which was mostly for me and sense-making. I thought about grabbing some JS templating tool like mustache, but found this [wes bos article](https://wesbos.com/template-strings-html/) and thought that native JS can handle this just fine.

### DIFF
--- a/app/views/catalog/_index_availability.html.erb
+++ b/app/views/catalog/_index_availability.html.erb
@@ -12,7 +12,9 @@
         </button>
       </p>
     </div>
-    <div class="availability-snippet col-md-9"></div>
+    <div class="availability-snippet col-md-9">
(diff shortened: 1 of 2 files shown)


app/views/catalog/_index_availability.html.erb
```diff
@@ -12,7 +12,9 @@
         </button>
       </p>
     </div>
-    <div class="availability-snippet col-md-9"></div>
+    <div class="availability-snippet col-md-9">
+      <p><span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span> Finding items...</p>
```
banukutlu: 👌

PR #317, opened 2019-05-20
changes based on upstream changes to the meta tags

cdmo:
see the changes at https://github.com/projectblacklight/blacklight/commits/master/app/views/layouts/blacklight/base.html.erb


cdmo:
wonder if there's anything else lurking in our views 🤔 

will poke around

banukutlu:
> wonder if there's anything else lurking in our views 🤔
> 
> will poke around

possible 🤔  yeah thanks!

Issue #227, opened 2019-02-14
Add catkey and link to librarian view to bottom of page


ruthtillman:
Similar to this: https://searchworks.stanford.edu/view/10321099

Should contain:

[View MARC record](https://blackcat01qa.libraries.psu.edu/catalog/2469352/librarian_view) | catkey: 2469352

ruthtillman:
Yay! 🎉

PR #326, opened 2019-05-23
marc view page title change

banukutlu:
part of #227


cdmo:
related, I really like this i18n feature of rails 👍

Issue #324, opened 2019-05-23
Rename "Back to Search"


ruthtillman:
First few testers were unsure if this restarted the process or returned us to search results. I sensed a little concern as well since we DON'T need the on-page "Back to Search" button to go back. The back button works! I agree it's not a helpful message as is.

Thoughts on whether it should be "Search Results," "Return to Search Results," or "Back to Search Results"?  Something else?

PR #329, opened 2019-05-23
rename back to search #324

banukutlu:
connected to #324


cdmo:
@banukutlu I approved this PR, but there was a conflict, so I resolved it, and now it won't let me merge it, I think you have to review my conflict resolution

Issue #325, opened 2019-05-23
Add primary author and edition to index results


ruthtillman:
For some reason, it looks like Edition didn't make it from [the list of fields](https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields#search-results-view-labels) to the [search results view](https://github.com/psu-libraries/psulib_blacklight/blob/da5c12c2bfb94161c08a007603763c54f2140242/app/controllers/catalog_controller.rb#L133-L138). That one should be easy to add.

Also add the primary author (editing the wiki to update).

Example of problem when there isn't enough info to give info about what the item is: https://blackcat01qa.libraries.psu.edu/?utf8=%E2%9C%93&search_field=title&q=american+marriage

ruthtillman:
Note -- changing this may change up the best place to put the format icon display... I'm open to either moving it around via CSS or [putting it at the bottom of the fields](https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=death+rituals).

PR #330, opened 2019-05-28
325 edition and authors to results display

banukutlu:
Connected to #325


spec/presenters/psul_index_presenter_spec.rb
```diff
@@ -22,15 +22,19 @@
...
+          config.add_index_field 'edition_display_ssm'
           config.add_index_field 'url_fulltext_display_ssm'
         end
       end
 
-      it 'has no label for publication statement and format but still has a label for URL' do
-        expect(presenter.fields['publication_display_ssm'].label).to be_nil
+      it 'has no label for format and publication statement but still has a label for URL' do
         expect(presenter.fields['format'].label).to be_nil
+        expect(presenter.fields['publication_display_ssm'].label).to be_nil
```
cdmo: Hmm, is there a reason to test `format` and `publication_display_ssm` to `be_nil` but not `author_person_display_ssm`, `author_corp_display_ssm`, `author_meeting_display_ssm`, and `edition_display_ssm`. I guess it seems like either test for one field or test for them all? 🤷‍♂️

banukutlu: There were only those fields on search results when I first wrote that spec so no specific reason. I can add the others  👍

banukutlu: Oh I remember now why not testing for others because the document does not actually have those fields.

cdmo: Ok, so I guess why didn't you mock these other fields too? Like, to me it makes sense to just mock one example of a field that shouldn't have a label and test that _or_ mock and test them all. Does that make sense?

Issue #319, opened 2019-05-20
Item view


cdmo:
Implementing mockup for individual item (moving links out of sidebar, removing sidebar, etc)

cdmo:
Assuming it's this [guy](https://psu.app.box.com/file/446221885088)

banukutlu:
> Assuming it's this [guy](https://psu.app.box.com/file/446221885088)

@ruthtillman ? for item view which mockup do we want to implement: the thumbnail left under or left beside one?

banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields#single-item-view-labels

ruthtillman:
yep that one.

cdmo:
The mockup drops back to search results and paging through results, which I assume we actually do want to keep. I'll just use my judgment and we can go from there.

cdmo:
Hi @ruthtillman, I'm about finished up with this issue as a "first step" kind of thing. Here's a screenshot example of an [item view (a first iteration)](https://user-images.githubusercontent.com/523381/58720762-1104a080-83a1-11e9-806d-1f2376d429f6.png). We'll deploy next week (after a review from @banukutlu) but while it's fresh in my mind I wanted to add a bit of context: 

1) The bookmark action is still just a checkbox input. I think I was wondering if we really wanted to keep the Bookmark feature or not. I think the answer is "we do" but thought I'd ask again just to make sure. If we do keep it, I can make the checkbox appear as a button (a la orangelight) 
2) I'm not sure about the display of secondary metadata. The mockup has it as a label + line break + contents at full width (or maybe it's not quite full width, I'll ask @mtribone ). [Text at full width on large screen is difficult to read](http://webtypography.net/2.1.2). This is to say that I've left it as the default two column layout for now (col 1 = label, col 2 = value).

mtribone:
@cdmo Yeah I definitely think that you can start getting into media queries and make some adjustments to the line length of the secondary metadata at full screen. Perhaps make some tests of 1200px and above and see how the text adjusts.

mtribone:
Maybe the "full screen" is for smaller resolutions and then 1200px and above gets centered at a certain width and has padding on the left and right.

cdmo:
Cool, thanks for the feedback. I also feel like it'd be useful to get holding availability details in here too. So it might be useful to sort of leave it where it's at until we get availability merged in (which should be soon)

[...]

PR #336, opened 2019-05-31
319 item view

cdmo:
WIP, meaning don't merge. But, I'd like a first review. What's missing is tests. Also missing is transforming the Bookmark action into a member of the record-level action link button bar.

Additional change: I moved the pivot facet sass out of psulib_icons.

Semi-related: I started using better_errors, it is _better_ than web_console. Thinking we should replace it in the Gemfile if you are okay with that (I'll add it to PR).

#319


app/views/catalog/_show_top_fields.html.erb
```diff
@@ -0,0 +1,16 @@
+<% document = @document %>
+<% doc_presenter = show_presenter(document) %>
+<dl class="row">
+  <% document_show_fields(document).each do |field_name, field| -%>
+    <% if render_top_field? document, field, field_name %>
+      <dt class="col-3 text-right blacklight-<%= field_name.parameterize %>-label"><%= render_document_show_field_label(document, field: field_name).html_safe %></dt>
+      <% if doc_presenter.configuration.show_fields.fetch(field_name)[:helper_method] %>
```
banukutlu: @cdmo I think we need a check for multi-valued fields for fields like additional creators like here: https://github.com/psu-libraries/psulib_blacklight/blob/master/app/views/catalog/_show_default.html.erb#L7?

banukutlu:
> WIP, meaning don't merge. But, I'd like a first review. What's missing is tests. Also missing is transforming the Bookmark action into a member of the record-level action link button bar.
> 
> Additional change: I moved the pivot facet sass out of psulib_icons.
> 
> Semi-related: I started using better_errors, it is _better_ than web_console. Thinking we should replace it in the Gemfile if you are okay with that (I'll add it to PR).
> 
> #319

@cdmo I think these changes looks good. A minor thing for multi valued fields for top section but also I know this is still WIP. But generally how you implemented the items view sections is looking good to me 👍

cdmo:
Cool, thanks @banukutlu. I'll return to working on it with this general approach then.

app/views/catalog/_show_top_fields.html.erb
cdmo: I refactored this a bunch. Can you [take a look when you have a chance](https://github.com/psu-libraries/psulib_blacklight/pull/336/commits/0c362e5c7ae16ee2b0bad896c91c3f2072e604a6)?

banukutlu: these changes looks great 👌 way cleaner 👍

PR #337, opened 2019-06-03
Announcement functionality and display

cdmo:
Just using application config

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -17,5 +17,8 @@ module Psulcat
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -17,5 +17,8 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
+    # When this is false, the value of the announcement bar is set in `blacklight.announcement.html`
```
banukutlu:
```suggestion
    # When this is true, the value of the announcement bar is set in `blacklight.announcement.html`
```

cdmo: Thanks for catching this 🤦‍♂

PR #338, opened 2019-06-03
Trying out displaying the links instead of a hamburger icon on mobile

cdmo:
this isn't tied to a specific issue, we don't really have a "mobile styling" issue, which I think is fine


cdmo:
Thanks!

Issue #196, opened 2018-12-06
Holdings and Availability Display


cdmo:
Holdings and summary holdings

https://github.com/psu-libraries/psulib_blacklight/wiki/Holdings-and-Availability-Display

banukutlu:
Connected to https://github.com/psu-libraries/psulib_traject/pull/61

banukutlu:
@ruthtillman 497977 has a holding with the below ItemInfo:
```
CallInfo>
<libraryID>ALTOONA</libraryID>
<classificationID>LC</classificationID>
<callNumber>BF76.7.P83 2010</callNumber>
<numberOfCopies>6</numberOfCopies>
...
<ItemInfo>
<itemID>000066834199</itemID>
<itemTypeID>BOOKFLOAT</itemTypeID>
<currentLocationID>ILLEND</currentLocationID>
<homeLocationID>STACKS-BR</homeLocationID>
<dueDate>2019-05-14T00:00:00.922-04:00</dueDate>
<chargeable>false</chargeable>
<numberOfHolds>1</numberOfHolds>
<fixedTimeBooking>false</fixedTimeBooking>
</ItemInfo>```

homeLocationID is Beaver - Stacks but currentLocationID is `"ILLEND": "This copy unavailable, submit request via Interlibrary Loan"` and LibraryID is ALTOONA. 

Since we are grouping by LibraryID, this holding shows up under Altoona:

![Screen Shot 2019-05-20 at 11 41 04 AM](https://user-images.githubusercontent.com/6477867/58037596-b98e4700-7afb-11e9-97b0-5f023f055aa4.png)

banukutlu:
@ruthtillman 52888 has a callInfo with `<libraryID>SERIAL-SRV</libraryID>`, SERIAL-SRV is not in the locations list that we were using for bento availability which we currently use while implementing BL availability. We talked about updating the locations list but also other than updating, what should we do in such cases if there is a libraryID that won't be in our list? Is this something we account for or not? Do we not list the holding and not count for available items?

ruthtillman:
I addressed the first part in the wiki. Re: `497977` -- considering.

ruthtillman:
Updated wiki in consultation with Banu.

cdmo:
[...]

PR #340, opened 2019-06-04
196 holdings and availability

banukutlu:
#196 

There are still couple minor things left to do (things waiting a reply/feedback from Ruth) but since it is mostly done, we think we can review and merge as is right now.


cdmo:
😤

Issue #110, opened 2018-10-02
Email/SMS Body fields


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields#emailsms-body

ruthtillman:
Reviewed fields. All Email/SMS fields are either indexed or being refined and are represented in tickets. Use and order documented in the wiki. We can close this ticket.

banukutlu:
after @cdmo pointed out, there is still work to do other than just the fields for this one.

cdmo:
Oh all I was saying is that we need *some* issue for SMS/Email. It can be this one, or this one can be closed and it can be some other new issue. The work that needs to be done is actually implementing this thing is all.

cdmo:
hi @ruthtillman, given our conversation in Teams a bit ago, I was wondering if we wanted to think again about SMS. Should we keep this functionality? Or because of the fact that since this functionality was originally developed, the vast majority of mobile phones became cameras and taking a picture of the computer screen is easier then having the record texted to you. I am on the fence myself.

I do think email should still be a feature for what it's worth.

ruthtillman:
What's the difficulty in setting up? I think at least in the long run, I'd like to have it... though if it's not working well out-of-the-box we can leave for now.

cdmo:
Haven't even looked at what's involved, I'd guess it's pretty simple. Just kinda had that doubt in my mind like "why even bother?" So I wanted to check with you. We'll throw it back in the backlog and get to it. Thanks!

banukutlu:
@ruthtillman do we want to add author field to Email/SMS as well now that we added it to search results page?

ruthtillman:
@banukutlu yes, let's add author.

ruthtillman:
Is this something I can test live? I tried using it but it may not be hooked up yet?

banukutlu:
> Is this something I can test live? I tried using it but it may not be hooked up yet?

@ruthtillman email should be working but sms not yet.

ruthtillman:
Email worked nicely for me! Includes link to record. Future work might be call numbers except this could potentially be a MESS with all our numbers and locations.

cdmo:
Author has been added, code deployed

PR #344, opened 2019-06-10
110 email sms fields

banukutlu:
Connected to #110

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -2,26 +2,29 @@
 
 class SolrDocument
   include Blacklight::Solr::Document
-  # The following shows how to setup this blacklight document to display marc documents
-  extension_parameters[:marc_source_field] = :marc_display_ss
-  extension_parameters[:marc_format_type] = :marcxml
-  use_extension(Blacklight::Solr::Document::Marc) do |document|
-    document.key?(:marc_display_ss)
-  end
-
-  field_semantics.merge!(
-    title: 'title_display_ssm',
-    language: 'language_facet_ssim',
-    format: 'format'
-  )
 
   # self.unique_key = 'id'
+  field_semantics.merge!(
+    title: 'title_display_ssm',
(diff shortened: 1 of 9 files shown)


app/models/solr_document.rb
```diff
@@ -2,27 +2,30 @@
 
 class SolrDocument
   include Blacklight::Solr::Document
-  # The following shows how to setup this blacklight document to display marc documents
```
cdmo: oh ha! I just noticed this in my latest PR on BL wondering if we could delete. Nice :)

app/models/solr_document.rb
```diff
@@ -2,27 +2,30 @@
...
-  extension_parameters[:marc_source_field] = :marc_display_ss
-  extension_parameters[:marc_format_type] = :marcxml
-  use_extension(Blacklight::Solr::Document::Marc) do |document|
-    document.key?(:marc_display_ss)
-  end
 
+  # self.unique_key = 'id'
```
cdmo: ?

app/models/solr_document.rb
cdmo: oh I see you just moved it down. We do need it then? Is this what produces the librarian view?

cdmo:
Nevermind, I see that Blacklight uses `deliver_now` or `deliver` not `deliver_later`!

banukutlu:
> Do you think we need specs for `to_sms_text` and `to_email_text`?

yup I was planning to write up some specs but is it OK if we add them later on, just so we can test emails on qa?

app/models/solr_document.rb
banukutlu: seems like we do not need it, it was already there I just changed the order of lines and kept it there since I was not sure 🤔

app/models/solr_document.rb
banukutlu: yeah I just moved it down

banukutlu: and yes I think it is for librarian view

PR #357, opened 2019-06-17
Item view fixes

cdmo:
Note that I have not done _much_ with fulltext link display. Pending conversation. But for the sake of the rest of these fixes moving on, putting in this PR.


app/views/catalog/_show_main_content.html.erb
```diff
@@ -17,9 +17,7 @@
 
   </div>
 
-  <div class="row document-metadata metadata-availability">
-    <%= render partial: 'show_availability', locals: { document: @document } %>
-  </div>
+  <%= render partial: 'show_availability', locals: { document: @document } if @document %>
```
banukutlu: yes moving that div to the partial makes so much cleaner 👍

PR #384, opened 2019-07-02
adding BKFLOATEMP location for availability

banukutlu:


banukutlu:
> is this supposed to be tied to an issue at all?

no there is no issue for it, I found that we did not have this location while testing on my local

PR #391, opened 2019-07-03
Issue 390 remap general stacks

ruthtillman:
Remapped general stacks name to Stacks - General Collection.


cdmo:
I jumped the gun on this one, but this is in production catalog now

PR #398, opened 2019-07-11
319 item view updates

cdmo:
for issue #319 

This should finish up this issue, though I'm sure we'll find some other things to work on 

Notice especially in that issue the check box to do items that have now been checked in the last few comments

## Additional change

* updated BL to a hash ref containing a fix important for dev work

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'rails', '~> 5.2.2'
 gem 'rsolr', '>= 1.0'
 gem 'webpacker', '~> 4.0'
 
-gem 'blacklight', '>= 7.0.1', github: 'projectblacklight/blacklight'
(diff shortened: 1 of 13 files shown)


Gemfile
```diff
@@ -20,7 +20,7 @@ gem 'rails', '~> 5.2.2'
 gem 'rsolr', '>= 1.0'
 gem 'webpacker', '~> 4.0'
 
-gem 'blacklight', '>= 7.0.1', github: 'projectblacklight/blacklight'
+gem 'blacklight', github: 'projectblacklight/blacklight', ref: '3e6cb292e0a7de606ec73b568292a6e170bc8cb6'
```
banukutlu: 👍

Issue #373, opened 2019-06-26
Bound-withs availability and holdings


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/wiki/Holdings-and-Availability-for-Bound-Withs

note: if there's lots of results, we'll need to: find ItemInfo WHERE BoundwithLinkInfo/ItemID == TitleInfo/CallInfo/ItemInfo/itemID to get the particular item to process

ruthtillman:
Works very well if only a child and/or if parent/child relationships. However, if the item is ONLY a parent, we're getting errors. Examples:

- https://blackcat01qa.libraries.psu.edu/catalog/1006984
- https://blackcat01qa.libraries.psu.edu/catalog/2794664

When searching for a parent, the whole search results page seems to glitch: https://blackcat01qa.libraries.psu.edu/?utf8=%E2%9C%93&search_field=title&q=The+federal+judiciary+%3A+a+Thanksgiving+discourse

Applies even when it's not the top result: https://blackcat01qa.libraries.psu.edu/?utf8=%E2%9C%93&f%5Bsubject_facet%5D%5B%5D=Judicial+power%E2%80%94United+States&search_field=dummy_range&range%5Bpub_date_itsi%5D%5Bbegin%5D=1860&range%5Bpub_date_itsi%5D%5Bend%5D=1880&commit=Limit

But a similar one I did before limiting by date is fine because we're not yet to the results page with that result on it: https://blackcat01qa.libraries.psu.edu/?f[subject_facet][]=Judicial+power%E2%80%94United+States

ruthtillman:
Looks good!

PR #405, opened 2019-07-19
373 bound withs availability

banukutlu:
Issue #373


app/javascript/psulib_blacklight/javascripts/availability.js
```diff
@@ -22,142 +28,239 @@ $(document).on('turbolinks:load', function() {
...
-    var allLocations = locations.locations;
-    var allLibraries = locations.libraries;
-    var illiadLocations = locations.request_via_ill;
-    var allItemTypes = item_types.item_types;
+function loadAvailability() {
     var titleIDs = [];
 
     // Get the catkeys
-    $('.availability').each(function() {
+    $('.availability').each(function () {
```
cdmo:
```suggestion
    $('.availability').each(function() {
```

Formatting, but I don't think we need a space between `function` and `()`. We should implement a linter.

Issue #422, opened 2019-07-31
Prevent Solr from searching from the homepage


cdmo:
#418 extension

PR #424, opened 2019-07-31
422 prevent solr

cdmo:
#422 

The `q` parameter is only present on intentional searches, not there on the homepage, where facets are only generated not results. So I used it's presence on whether or not to hit Solr.

Also added pry-byebug for making debugging easier.


app/controllers/catalog_controller.rb
```diff
@@ -6,6 +6,13 @@ class CatalogController < ApplicationController
   include Blacklight::Catalog
   include Blacklight::Marc::Catalog
 
+  # Only get search results from the solr index
+  def index
+    return nil unless params.key? 'q'
```
banukutlu: nice 👌

PR #426, opened 2019-08-01
Permits searches with no q but only an f specified to hit Solr

cdmo:


app/controllers/catalog_controller.rb
```diff
@@ -8,7 +8,7 @@ class CatalogController < ApplicationController
 
   # Only get search results from the solr index
   def index
-    return nil unless params.key? 'q'
+    return nil unless params.key?('f') || params.key?('q')
```
banukutlu: good catch 👌

PR #436, opened 2019-08-08
Changing special collections location text

ruthtillman:
Changed labels for SCL/Archives locations. This is a temporary fix before Aeon deployment. (I think I got it all)


ruthtillman:
Looks good!

Issue #430, opened 2019-08-02
Deep paging


cdmo:
see https://github.com/Brown-University-Library/bul-search/blob/master/app/controllers/catalog_controller.rb#L503-L513

cdmo:
considering this a dupe of #342

PR #439, opened 2019-08-08
430 deep paging

cdmo:
#430


app/controllers/catalog_controller.rb
```diff
@@ -342,4 +342,24 @@ def index
 
     config.raw_endpoint.enabled = true
   end
+
+  PAGINATION_THRESHOLD = 250
+  before_action only: :index do
+    if params[:page] && params[:page].to_i > PAGINATION_THRESHOLD
+      Rails.logger.info("Pagination threshold exceeded for #{request.ip} (#{request.user_agent}). Params: #{params}")
```
banukutlu: 👌 nice to log these!

PR #448, opened 2019-08-21
429 refactor suggestions

cdmo:


app/javascript/psulib_blacklight/javascripts/availability.js
```diff
@@ -383,29 +379,67 @@ function groupByLibrary(holdings) {
...
+                                data-item-type="${item.itemTypeID}"
+                                data-item-id="${item.itemID}"
+                                data-item-location="${aeonLocationText}"
+                                href="#">${spinner}View in Special Collections</a>`;
+
+        return aeonLocation;
+    }
+
+    // Special Colletions
+    if (['UP-SPECCOL'].includes(item.libraryID)) {
```
banukutlu: @cdmo We need to use the Archival Thesis links for for locations `'ARKTHESES', 'AH-X-TRANS'`, since there might be such location for the `UP-SPECCOL` library, we should either check if location is not one of the archival thesis ones here or we do it in a else if block or change the order of the if blocks?

cdmo: So, https://github.com/psu-libraries/psulib_blacklight/pull/448/files#diff-ae31f8e714d91a5996b94e9240f9256eR398-R423 (lines 398 to 423) are always going to be executed first.

The code in `429-aeon-links-non-thesis` has logic that treats this as an OR as well, i.e, [an else if followed by an else if](https://github.com/psu-libraries/psulib_blacklight/blob/429-aeon-links-non-thesis/app/javascript/psulib_blacklight/javascripts/availability.js#L393-L403). If the first scenario is met on your branch (i.e., ['ARKTHESES', 'AH-X-TRANS'] is present) it won't get to the next scenario regarding ['UP-SPECCOL'] right?

banukutlu: > If the first scenario is met on your branch (i.e., ['ARKTHESES', 'AH-X-TRANS'] is present) it won't get to the next scenario regarding ['UP-SPECCOL'] right?

correct ☝️

cdmo: > We need to use the Archival Thesis links for for locations 'ARKTHESES', 'AH-X-TRANS'

That's what [this](https://github.com/psu-libraries/psulib_blacklight/pull/448/files#diff-ae31f8e714d91a5996b94e9240f9256eR399) (line 399) is doing though

app/javascript/psulib_blacklight/javascripts/availability.js
```diff
@@ -10,6 +10,7 @@ var allLocations = locations.locations;
 var allLibraries = locations.libraries;
 var illiadLocations = locations.request_via_ill;
 var allItemTypes = item_types.item_types;
+const scanURL = 'https://psu.illiad.oclc.org/illiad/upm/lending/lendinglogon.html';
```
banukutlu: @cdmo `scanURL` seems to be not used, maybe intended to be used in line 410: https://github.com/psu-libraries/psulib_blacklight/blob/9bfe687facb564fe268bcb22b38cad5918ee05b7/app/javascript/psulib_blacklight/javascripts/availability.js#L410

app/javascript/psulib_blacklight/javascripts/availability.js
banukutlu: Yes, it does but for holdings with `UP-SPECCOL` as library and `['ARKTHESES', 'AH-X-TRANS']` as currentLocationID, [this](https://github.com/psu-libraries/psulib_blacklight/pull/448/files#diff-ae31f8e714d91a5996b94e9240f9256eR426) (line 426) should not be true, it should only do line `399`. So sth like this
`if (['UP-SPECCOL'].includes(item.libraryID) && !(['ARKTHESES', 'AH-X-TRANS'].includes(item.locationID))) {` should address those cases, or another way to filter out those cases.

app/javascript/psulib_blacklight/javascripts/availability.js
cdmo: Yes. My mistake. Sorry! Gonna close this PR, it's a mess.

app/javascript/psulib_blacklight/javascripts/availability.js
banukutlu: per our convo with @cdmo this is correct, it should work correctly since once it finds an archival thesis, it returns. I did not see that return originally.

cdmo:
Phew. This was a little messy. :)

Issue #444, opened 2019-08-19
fix mobile header problem


cdmo:
![Image-1](https://user-images.githubusercontent.com/523381/63265778-604dc500-c25c-11e9-92b0-c78bd56c5424.jpg)

header appears collapsed/hidden on mobile

cdmo:
As a footnote to this issue, the screenshot above is from an iPhone SE. This issue doesn't occur on Firefox emulating iPhone SE though. Nor does it appear when resizing the browser window to a narrow viewport on Firefox either. However, opening in Safari and narrowing the window does show.

I did check with someone with an iPhone 7 and the problem was present there too.

cdmo:
this was fixed, neglected to close

PR #449, opened 2019-08-21
444 header problem

cdmo:
#444 

## Additional fixes:

Removed a redundant import of header partial  
Got rid of blacklight-main-container as a part of `container_classes` because it was getting applied in more places than I realized (responsible for adding margin to the navbar which wasn't desired)


app/helpers/layout_helper.rb
```diff
@@ -21,6 +21,6 @@ def sidebar_classes
   # overwritten to return 'container-fluid' for Bootstrap full-width layout
   # @return [String]
   def container_classes
-    'container main-blacklight-container'
```
banukutlu: @cdmo do we still need [this](https://github.com/psu-libraries/psulib_blacklight/blob/ab0a9cfc8a6692cd01aa933ec9b583ca178a7136/app/javascript/psulib_blacklight/styles/_mobile.scss#L34) after this change?

app/javascript/psulib_blacklight/styles/index.scss
```diff
@@ -4,7 +4,8 @@
 
 // Twitter bootstrap
 @import '~bootstrap/scss/bootstrap';
-@import 'header';
```
banukutlu: 👍

app/helpers/layout_helper.rb
cdmo: Oh shoot, I meant to just remove `main-blacklight-container` thank you for catching. But, yeah, we could just kill the whole class because it's identical to it's parent.

app/javascript/psulib_blacklight/styles/index.scss
cdmo: more like 🤦‍♂

app/views/layouts/blacklight/base.html.erb
```diff
@@ -39,7 +39,7 @@
 <%= render partial: 'shared/announcement' if Psulcat::Application.config.show_announcement %>
 <%= render partial: 'shared/header_navbar' %>
 
-<div id="main-container" class="<%= container_classes %>">
+<div id="main-container" class="blacklight-main-container <%= container_classes %>">
```
banukutlu: @cdmo just trying to understand, what is `blacklight-main-container` for?

cdmo: Good question. It's used in [_mobile.scss](https://github.com/psu-libraries/psulib_blacklight/blob/ab0a9cfc8a6692cd01aa933ec9b583ca178a7136/app/javascript/psulib_blacklight/styles/_mobile.scss#L34-L37)

Just to add margin to the top and bottom at smaller breakpoints because otherwise the container slams up into the header.

I mistakenly thought in a previous commit that it was only applying to the main central container, but it applies to the navbar container and search form container too so it was creating extra spacing where not intended and i dont know how I missed it before. See areas in yellow boxes

Screenshot 2019-08-22 09 54 58

cdmo: Oh and I generally avoid using `id`s in CSS as a rule. Here's [a good post on why](http://oli.jp/2011/ids/)

Main reason is brittle-ness and just the general feeling that `id`s are redundant in CSS. Honestly, I don't like them for anything. lol

cdmo:
Thanks for the catches!

PR #453, opened 2019-08-22
bookmarks page - add share btn add and remove sort widget

banukutlu:
Issue #447 and Issue #452

### DIFF
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class BookmarksController < CatalogController
+  include Blacklight::Bookmarks
+
+  configure_blacklight do |_config|
+    _config.index.collection_actions.delete(:sort_widget)
+  end
+end
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -3,7 +3,7 @@
 class BookmarksController < CatalogController
   include Blacklight::Bookmarks
 
-  configure_blacklight do |_config|
+  configure_blacklight do |config|
     _config.index.collection_actions.delete(:sort_widget)
   end
 end
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -4,6 +4,6 @@ class BookmarksController < CatalogController
   include Blacklight::Bookmarks
 
   configure_blacklight do |config|
-    _config.index.collection_actions.delete(:sort_widget)
+    config.index.collection_actions.delete(:sort_widget)
   end
 end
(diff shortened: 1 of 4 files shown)


cdmo:
I see that my review has been requested, but, I see "WIP" so, I'm guessing you don't want a review yet?

app/views/catalog/_sort_widget.html.erb
```diff
@@ -0,0 +1,13 @@
+<% if show_sort_and_per_page? and active_sort_fields.many? %>
```
cdmo: So this "tidying sort" for general results isn't in either issue, but it's an additional change kind of thing right? This is mainly to make the buttons for sort and "per page" full width on mobile? Was someone asking for this change?

app/controllers/bookmarks_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class BookmarksController < CatalogController
+  include Blacklight::Bookmarks
+
+  configure_blacklight do |_config|
+    _config.index.collection_actions.delete(:sort_widget)
```
cdmo:
```suggestion
    config.index.collection_actions.delete(:sort_widget)
```

app/controllers/bookmarks_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class BookmarksController < CatalogController
+  include Blacklight::Bookmarks
+
+  configure_blacklight do |_config|
```
cdmo:
```suggestion
  configure_blacklight do |config|
```

An underscore before a var is, I believe, only used on vars that don't get called. Right?

app/views/bookmarks/_tools.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="share-widget" role="group" aria-label="Actions including with nested dropdown">
```
cdmo: so this is taking part of `_record_toolbar.html.erb` right?

app/views/catalog/_sort_widget.html.erb
banukutlu: Right, this is kinda of an additional change. Since I changed the share widget to full width on mobile, I ended up changing per page button as well. Then since I changed the per page, it is also changed on search results as well so I thought I would change the sort to be full width so all are consistent.

app/views/bookmarks/_tools.html.erb
banukutlu: correct, and I added  `data-display="static"` to line 3 and `dropdown-menu-right dropdown-menu-md-left` classes to line 4.

app/views/catalog/_sort_widget.html.erb
cdmo: Yup, makes sense. And I know you showed it to me in a demo, I think anyway, but, I didn't realize until I saw the code.

PR #465, opened 2019-09-10
Js linting

cdmo:
Let's figure out what to do with `233-jest` first


cdmo:
```
➜  psulib_blacklight git:(js-linting) yarn lint
...
✖ 118 problems (118 errors, 0 warnings)
```
😰

cdmo:
Most of the problems were either undefined vars that are globals (jquery and blacklight) and max length violations.

Playing around with this some, there is an additional "cyclomatic complexity" cop (err, I mean rule). See https://eslint.org/docs/rules/complexity. When I turn this on and set the level to 2 we do encounter several violations. I kinda like this rule.

Also, we can run a "Fix" with this tool too, though I haven't gotten it to do much:

`node_modules/.bin/eslint --fix --ext .js app/javascript spec/javascript`

Probably because the errors we had couldn't be fixed automatically.

FWIW: there is no "generate todo" kinda feature, see https://github.com/eslint/eslint/issues/9844

app/javascript/psulib_blacklight/javascripts/availability.js
```diff
@@ -349,7 +361,7 @@ function availabilityDataStructurer(holdingMetadata) {
 
             // Supplement data with an itemType and remove callNumber conditionally
             holdingMetadata[libraryID].forEach(function(element) {
-                element.itemType = (element.itemTypeID in allItemTypes) ? allItemTypes[element.itemTypeID] : "";
+                element.itemType = (element.itemTypeID in allItemTypes) ? element.itemTypeID : "";
```
banukutlu: @cdmo shouldn't this still be `allItemTypes[element.itemTypeID]`?

cdmo: Good catch, yes, sorry for the mistake

banukutlu:
```suggestion
                element.itemType = (element.itemTypeID in allItemTypes) ? allItemTypes[element.itemTypeID] : "";
```
per our conversation with @cdmo, reverting this back

Issue #233, opened 2019-03-06
Feature tests


cdmo:
Create feature tests for:
- [x] Title, author, subject dropdown searches
- [x] Advanced search
- [x] _search_results.html.erb 
- [x] Availability - handled with #450 
- [x] Subjects
- [x] Genre
- [x] Publication fields

Create tests for our behaviors (javascript) to make sure things don't break when we update javascript code or  JS dependencies (like Bootstrap)

banukutlu:
Make sure to cover https://github.com/psu-libraries/psulib_blacklight/wiki/Regression-Testing and https://github.com/psu-libraries/psulib_blacklight/wiki/Acceptance-Tests

cdmo:
> Create some tests for our behaviors (javascript) to make sure things don't break when we update javascript code or js dependencies (like Bootstrap)

Well, I did this part of the issue for book covers using `jest` and importing fixtured data during CI. It tests if the JavaScript is functioning at a unit level (jest) and then makes sure the actual image from Google Books shows up on the page by using a selenium javascript-enabled chromedriver test. I made this a feature test even though there aren't any clicks or interactions. Think it's okay because the implied integration is the JS working behind the scenes to do the JSONP "AJAX" work and replacement of the tag. So that's one thing anyway.

Wonder if we should make break apart the Regression Testing and Acceptance Tests wiki pages into new issues. Also, we should talk about using Cucumber versus just representing acceptance tests in rspec capybara feature tests. Do we want the cucumber overhead, I dont know.

PR #467, opened 2019-09-13
233 feature tests

cdmo:
#233


cdmo:
Without adding [puffing billy](https://stackoverflow.com/a/17007324/721065), we can't stub JSONP. I am thinking that we just not worry about stubbing for Google Book API's book covers. Probably not worth the complexity of adding another gem and configuring.

SO this PR has no actual mocking after all lol

cdmo:
For this particular spec to work, we'd also need to figure out a way to load Solr fixture data

cdmo:
Ok, updated the fixtures and made CI (Travis) run an index on that, then clear it at the end (not sure that's really necessary though). Think this is ready to have WIP removed.

spec/features/book_covers_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.feature 'BookCovers', type: :feature do
+  describe 'User searches for a book' do
+    it 'a book that happens to have a book cover in the Google Books API', js: true do
```
banukutlu:
```suggestion
    it 'that happens to have a book cover in the Google Books API', js: true do
```

banukutlu: minor update ☝️

Issue #480, opened 2019-10-11
Temporary: I Want It Button won't work right during upgrade


ruthtillman:
During the upgrade, the "I Want It" button won't work properly. It may redirect people to Test Cat, which will fail to put the item on hold. It may load an error. We aren't sure. How should we deal with this?

Complications:

1. The problem will only exist during the upgrade, approx midnight to noon on Oct 19th.
2. The button must otherwise be usable because it's a core part of our catalog user experience

Because this is a fairly limited problem, we could:

1. Allow a poor user experience during this low-traffic time and hope people understand from the other warnings that this is just a temporary issue 
2. Remove it that night and redeploy on Saturday afternoon
3. Is there a time-based test we temporarily add to determine whether it displays? I know how to do this in PHP. It's not _good_ but would avoid poor user experience and not require an immediate redeploy.

@cdmo @banukutlu I'm leaving this mostly to your judgment.

cdmo:
Thanks @ruthtillman - do we know what the HTTP response will be from the I Want It services? Will it `500`? Or will it `200`?

ruthtillman:
@cdmo I do not know. I believe that Maryam and Todd may have more info on that.

cdmo:
@mkutch can you respond to my question above?

tdouty:
I'm not sure what it will be. Are you using port 28443 or 9443?

tdouty:
9443 is apache http  and 28443 is Tomcat.

cdmo:
Both.

banukutlu:
> 9443 is apache http and 28443 is Tomcat.

talking about ports, which one we should be using? We have `28443` but for items info we do another call with `9443`.

cdmo:
So if I'm reading Ruth's note correctly you're not sure if it will redirect to test-cat? Essentially you don't know what will happen during the outage. Is that right?

The port stuff is for availability lookup info - which is a different issue, right?

ruthtillman:
Seems like convos on here also engage with the question in #479. ... IF we can have someone available on the weekend to load BlackCat on Friday about 8pm with the message and availability turned off and then turn things back on back on Saturday when it's back, then we don't need to build in these different components.

[...]

Issue #479, opened 2019-10-11
Temporarily change Availability error message (deploy Friday 10/18?)


ruthtillman:
During the upgrade on October 19th, the entire holdings/availability section won't load. During that period, the error message should be specific to the reason, rather than present it as a generic error.

Update
https://github.com/psu-libraries/psulib_blacklight/blob/2ab7b1bfd65551818121595a58a3bf7c27199cc7/app/javascript/psulib_blacklight/javascripts/availability.js#L548-L555

With the error message text:

```Required system maintenance is currently being performed. Availability and I Want It functions should be restored by noon October 19th. [See library alerts](https://libraries.psu.edu/alerts) for updates and details.```

We'll then need to revert this to the old error after the maintenance has been completed.

cdmo:
How about we disable holdings and add just one error message to the top of the page? I just think it'd be better than displaying the same error so many times on the page and through the experience in the application. Thoughts?

banukutlu:
> How about we disable holdings and add just one error message to the top of the page? I just think it'd be better than displaying the same error so many times on the page and through the experience in the application. Thoughts?

I liked this approach very much, what do you think @ruthtillman ?

ruthtillman:
@banukutlu This is doable but only if we have someone, probably you or @cdmo, prepared to restore Blacklight as soon as that the system is working again (10am-noon Saturday). Because we can't just leave Availability turned off all weekend. The downside is the amount of work it would ask you to do outside of normal working hours. Thoughts? I may be available to chat details via Slack on and off today.

banukutlu:
@ruthtillman sounds like you are fine with displaying one msg on top, cool 👍 
for the rest, we are thinking about our approach, we will let you know.

banukutlu:
We will not display `Availability` and `I want It` during read-only and display the error msg on top under the search bar rather than displaying an error msg for each record.

PR #483, opened 2019-10-17
readonly status

banukutlu:
Issue #479 and #480 

thanks @cdmo!


.rubocop_todo.yml
```diff
@@ -9,12 +9,12 @@
 # Offense count: 1
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/BlockLength:
-  Max: 42
+  Max: 28
```
cdmo: Went down? 🤔

app/controllers/catalog_controller.rb
```diff
@@ -360,4 +362,22 @@ def index
...
+      @readonly_settings['read_only']
+    else
+      false
+    end
+  end
+
+  private
+
+    def readonly_message
+      flash.now[:error] = ActionController::Base.helpers.sanitize(@readonly_settings['message']) if readonly?
```
cdmo: Happy to call a class method instead of a mysterious variable 👍

Issue #292, opened 2019-05-07
Remove search on top from the advanced search page


banukutlu:

banukutlu:
@ruthtillman I just checked some of the other Advanced Search implementations like Stanford, Princeton and Temple, they kept the search on top. What do you think, should we remove or keep it?

ruthtillman:
Per our discussion, let's remove the whole bar!

ruthtillman:
Just wanted to reinforce that I've seen this cause momentary confusion in some tests as well. The button saying "Search" makes people reviewing their terms think they can click "Search" at the top and have it perform the advanced search.

PR #492, opened 2019-10-24
hide searchbar on AS page

banukutlu:
Issue #292

### DIFF
--- a/app/views/shared/_header_navbar.html.erb
+++ b/app/views/shared/_header_navbar.html.erb
@@ -8,8 +8,10 @@
   </div>
 </nav>
 
+<% unless @search_fields_for_advanced_search %>
 <div class="navbar-search navbar navbar-expand-md navbar-light bg-light" role="navigation">
   <div class="<%= container_classes %>">
-    <%= render_search_bar %>
+      <%= render_search_bar %>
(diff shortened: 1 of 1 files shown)


spec/views/shared/_header_navbar.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    assign(:search_bar, Blacklight::SearchBarPresenter.new(controller, blacklight_config))
+  end
+
+  it 'displays search bar except on Advanced Search' do
+    render 'shared/header_navbar'
+    expect(rendered).to have_selector '.navbar-search'
+  end
+
+  it 'does not display search bar on Advanced Search' do
+    assign(:search_fields_for_advanced_search, Blacklight::SearchBarPresenter.new(controller, blacklight_config))
```
cdmo: What do you think of this? Since you already are `assign`ing `Blacklight::SearchBarPresenter.new(controller, blacklight_config))` above?

```suggestion
    assign(:search_fields_for_advanced_search, :search_bar)
```

banukutlu: yes, good idea 👍

PR #500, opened 2019-10-29
Fix dummy range bug

cdmo:
Had to do a bit of refactoring of the helper to pass :cop:, especially `range_display`


cdmo:
yes.
I'll change.

cdmo:
tabling this for another time, maybe it'll get fixed upstream

PR #511, opened 2019-11-07
On demand sitemap

cdmo:
Inspired by https://github.com/sul-dlss/SearchWorks/pull/2351


cdmo:
When this is reviewed, I'd like to deploy into QA without indexing. Then I'll send a manual update commit. I'm not sure if a full index is required.

solr/conf/solrconfig.xml
```diff
@@ -56,17 +56,22 @@
 
   </updateHandler>
 
-  <updateRequestProcessorChain name="cloud" default="true">
-    <processor class="solr.LogUpdateProcessorFactory"/>
-    <processor class="solr.DistributedUpdateProcessorFactory"/>
```
banukutlu: > Solr will automatically insert DistributedUpdateProcessorFactory in any chain that does not include it just prior to the RunUpdateProcessorFactory

well I did not know this ☝️, now learned :)

cdmo: Actually I think the removal of this `updateRequestProcessorChain` was a holdover from some trial and error I was doing. I am going to add this back in. Thanks for catching.

Now I'm wondering if maybe we should explicitly add it to the dedupe `updateRequestProcessorChain`.

cdmo: Ok, I've read over the refguide (especially [Configuring Individual Processors as Top-Level Plugins](https://lucene.apache.org/solr/guide/7_4/update-request-processors.html#configuring-individual-processors-as-top-level-plugins))and the [solr generated code docs](https://lucene.apache.org/solr/7_4_0//solr-core/org/apache/solr/update/processor/UpdateRequestProcessorChain.html) on URPs and this is where I landed. There are alternatives, but I think this is succinct and clear enough.

Kinda second-guessing adding a simple hash `to_field` in traject, but 🤷‍♂

PR #524, opened 2019-12-04
updating failing specs

banukutlu:


spec/features/availability_spec.rb
```diff
@@ -82,6 +82,7 @@
   describe 'Archival Material:' do
     it 'has a \'Request Material\' link so it can be requested through Aeon', js: true do
       visit '/catalog/1836205'
+      click_button('View More')
```
cdmo: Is this necessary? Box 1 the first one in the first 4 listed...

spec/features/availability_spec.rb
```diff
@@ -65,13 +65,13 @@
   describe 'User opens a record with many holdings in a library' do
     it 'toggles 4 or more holdings with View More/View Less button', js: true do
       visit '/catalog/1793712'
-      within 'div[data-library="UP-MICRO"]' do
+      within 'div[data-library="UP-PAT"]' do
```
cdmo: Curious. Did we get rid of our micro collection then?

banukutlu: not sure if we got rid of that collection but this record does not have UP-MICRO anymore

spec/features/availability_spec.rb
banukutlu: The reason I added that click is to make sure that the holding we are testing here would always be in the visible css. With being not 💯 sure, I think this is the spec that is failing time to time depending on the order of holdings listed, so I thought adding a click to view more would work around that.

cdmo: Ok, cool. Thanks.

PR #558, opened 2020-04-04
Adds config gem to replace Blackcat Messages/Admin options.

cdmo:
This fully replaces blackcat admin/messages. I was looking at your PR and it was just after I was playing with figaro to work on disabling masquerading in production (with WebAccess) and I just couldn't help but give it a try. So you can see it's a lot less code. One catch is that the server has to be restarted when ENV values change. I kinda like this approach better, let me know what you think?


app/views/catalog/_show_main_content.html.erb
```diff
@@ -17,7 +17,7 @@
 
   </div>
 
-  <%= render partial: 'show_availability', locals: { document: @document } if @document && !blackcat_config.blackcat_message?(:readonly) %>
+  <%= render partial: 'show_availability', locals: { document: @document } if @document unless ENV['readonly'] %>
```
banukutlu, line 20: `if @document unless ENV['readonly']` would this work 🤔

cdmo: It did work in my testing, were you able to test?

app/views/shared/_announcement.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="announcement container-fluid bg-light">
   <div class="row align-items-center">
     <div class="col align-content-center">
-      <p class="text-center mt-2 mb-2 pb-1"><i class="fas fa-bullhorn"></i><%= blackcat_config.blackcat_message(:announcement) %></p>
+      <p class="text-center mt-2 mb-2 pb-1"><i class="fas fa-bullhorn"></i> <%= ENV['announcement'] || t('blacklight.announcement.html') %></p>
```
banukutlu, line 4:
```suggestion
      <p class="text-center mt-2 mb-2 pb-1"><i class="fas fa-bullhorn"></i> <%= sanitize ENV['announcement'] || t('blacklight.announcement.html') %></p>
```

banukutlu: above change because 

[](url)
Screen Shot 2020-04-08 at 1 10 56 PM

app/controllers/application_controller.rb
```diff
@@ -14,19 +14,11 @@ class ApplicationController < ActionController::Base
...
 
   private
 
-    def flash_readonly
-      flash.now[:error] = blackcat_config.blackcat_message(:readonly) if blackcat_config.blackcat_message?(:readonly)
-    end
-
     def flash_alert
-      flash.now[:error] = blackcat_config.blackcat_message(:alert) if blackcat_config.blackcat_message?(:alert)
+      flash.now[:error] = ENV['alert']
```
banukutlu, line 22:
```suggestion
      flash.now[:error] = sanitize ENV['alert']
```

banukutlu, line 22: same here, needs to be sanitized

maybe this?

```suggestion
      flash.now[:error] = Rails::Html::LinkSanitizer.new.sanitize(ENV['alert'])
```

app/controllers/application_controller.rb
```diff
@@ -14,19 +14,11 @@ class ApplicationController < ActionController::Base
...
-  before_action :flash_readonly, :flash_alert
-
-  def blackcat_config
-    @blackcat_config ||= BlackcatConfig::Builder.new
-  end
+  before_action :flash_alert
 
   private
 
-    def flash_readonly
```
banukutlu, line 25: @cdmo how does the readonly flash will be displayed?

banukutlu: is the idea set the `readonly` to true and also set the `alert`? that would work I think, and liked this idea

banukutlu: But I could not get it to work like that 🤔

app/views/catalog/_show_main_content.html.erb
banukutlu: it worked yup ✔️

app/controllers/application_controller.rb
cdmo: yup, basically just make one channel for setting that. However, I have a new thought based on #559. I think we can just use the announcement bar. Like let's just replace that when there is an alert. Obviously setting flash :error seems to have unintended side affects.

cdmo:
Using config gem instead of figaro because we're already using confing in my account and because you can use ruby objects (hash, string, bool, etc) and because figaro seems like it's best used for storing secrets/system-level configuration (keys, connection strings, etc)

app/views/shared/_announcement.html.erb
cdmo: Thanks, this is resolved in latest

app/controllers/application_controller.rb
cdmo: got rid of flash alert entirely 👌

app/controllers/application_controller.rb
cdmo: gone now due to #559

cdmo:
So this is merging into `554` branch. When this merges, I guess I have to merge 554 into master. Just a note.

PR #560, opened 2020-04-13
rails upgrade to address actionview vulnerability

banukutlu:
To address [this](https://github.com/psu-libraries/psulib_blacklight/network/alert/Gemfile.lock/actionview/open)

## Additional Work
Update eslint to address the [acorn vulnerability](https://github.com/psu-libraries/psulib_blacklight/network/alert/yarn.lock/acorn/open)


cdmo:
We don't use the escape literal, so this isn't urgent, right? The acorn and es lint ones are build-time, so also probably not urgent. Let's try and get blackcat admin changes merged in first? I see the failing spec for alert (which is kinda odd, but also irrelevant since that is going away)

PR #563, opened 2020-04-15
addresses minimist and serialize-javascript vulnerabilities

banukutlu:


banukutlu:
> we had to upgrade webpacker to fix these?

yeah, a bit more about it here https://github.com/rails/webpacker/issues/2387

banukutlu:
@cdmo this is ready for review finally :)

PR #564, opened 2020-04-17
Fix for INC1422412, wrong image showing when ISBN set goes first

cdmo:


cdmo:
verified that this is working locally for the book in question. Thanks to @mkutch for the catalog dump. FYI @ruthtillman we are adjusting our formula to get book covers from google to match the cat: oclc nums first (we were sending isbns first)

cdmo:
Screen Shot 2020-04-17 at 2 20 39 PM

banukutlu:
minor change required for the spec, can be merged after the change

PR #566, opened 2020-05-06
Jquery fix

cdmo:
Additionally:

* removed sass-loader (not needed)
* updated blacklight-frontend to match our installed blacklight


spec/features/availability_spec.rb
```diff
@@ -86,7 +86,7 @@
   end
 
   describe 'Archival Material:' do
-    it 'has a \'Request Material\' link so it can be requested through Aeon', js: true do
+    it 'has a "Request Material" link so it can be requested through Aeon', js: true do
```
cdmo, line 89: oh yeah I changed this too, just because I saw it and it seemed like an unneeded escaping

package.json
```diff
@@ -4,16 +4,15 @@
   "dependencies": {
     "@fortawesome/fontawesome-free": "^5.7.2",
     "@rails/webpacker": "~4.0",
-    "blacklight-frontend": "^7.0",
+    "blacklight-frontend": "7.5.0",
```
banukutlu, line 7: 👌

PR #574, opened 2020-05-18
Removes a few local customizations from the blacklight-override group

cdmo:
Just getting this in here before I forget, this was discovered during Refactor Friday session

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ env:
 addons:
   chrome: stable
 
+branches:
+  only:
+    - master
+
 before_install:
   - yes | gem update --system
   - gem install bundler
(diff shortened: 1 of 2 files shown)


cdmo:
Additionally a config for travis to mirror what we did in MyAccount - basically it'll run CI on feature branch pushes if they are PRs as well as pushes to master (when we merge PRs in). Also I changed travis settings (from the web) to run on "build pushed branches" and "build pushed prs" (instead of just on "build pushed PRs"). Telling travis to only work on branches: `master` through `.travis.yml` means that pushed branches will only initiate builds on `master` not feature branches, but that prs will still initiate builds because of the "build pushed prs" setting. 

Without these changes PRs would get two builds and master would never get built either which means no report on test coverage.

PR #584, opened 2020-06-03
Metadata display fixup

cdmo:


banukutlu:
Screen Shot 2020-06-04 at 2 19 40 AM

@cdmo Looks like it is not aligned with the above metadata on my local, same for you?

cdmo:
Yeah I saw that. Thought it was supposed to be that way? I honestly don't know. Blacklight styling right-aligns this by default. But, I agree that it looks funky. Think we should left align it?

cdmo:
Flexbox stuff acts in a weird way when it has to position around a `block` that is `float`ed right. It seems to add extra spacing on left, and then when that `block` (the thumbnail) is above and not affecting layout, the extra spacing on left goes away.

Issue #470, opened 2019-09-18
Bento integration for Blacklight


cdmo:
https://git.psu.edu/i-tech/ul/-/issues/681

Aha! Link: https://penn-state-university-libraries.aha.io/features/STPROJ-23

PR #588, opened 2020-06-05
Bento Controller

cdmo:
* Also correctly specifies the requestHandler

For JSON requests, removes facets and the included information

#470 

before:

`1]  webapp=/solr path=/select params={facet.field=access_facet&facet.field=format&facet.field={!ex%3Dcampus_facet_single}campus_facet&facet.field={!ex%3Dlibrary_facet_single}library_facet&facet.field={!ex%3Dup_library_facet_single}up_library_facet&facet.field=pub_date_itsi&facet.field=language_facet&facet.field=subject_topic_facet&facet.field=genre_facet&facet.field=media_type_facet&facet.field=lc_1letter_facet&facet.field=lc_rest_facet&f.library_facet.facet.sort=index&facet.pivot=lc_1letter_facet,lc_rest_facet&f.library_facet.facet.limit=-1&f.up_library_facet.facet.sort=index&sort=score+desc,+pub_date_itsi+desc,+title_sort+asc&rows=3&f.campus_facet.facet.limit=-1&f.lc_rest_facet.facet.sort=index&f.subject_topic_facet.facet.limit=21&q=green&defType=edismax&stats=true&f.lc_1letter_facet.facet.sort=index&f.up_library_facet.facet.limit=-1&f.media_type_facet.facet.limit=21&f.genre_facet.facet.limit=21&f.campus_facet.facet.sort=index&facet=true&wt=json&f.language_facet.facet.limit=11&stats.field=pub_date_itsi} hits=7 status=0 QTime=2
2020-06-05 19:44:47.451 INFO  (qtp1095352419-67) [c:psul_blacklight s:shard1 r:core_node2 x:psul_blacklight_shard1_replica_n1] o.a.s.c.S.Request [psul_blacklight_shard1_replica_n`

after:

`1]  webapp=/solr path=/search params={q=green&defType=edismax&stats=true&sort=score+desc,+pub_date_itsi+desc,+title_sort+asc&rows=3&facet=false&wt=json&stats.field=pub_date_itsi} hits=7 status=0 QTime=1`

left to do:

- [x] Figure out specs


app/controllers/catalog_controller.rb
```diff
@@ -64,7 +64,7 @@ def index
     }
 
     # solr path which will be added to solr base url before the other solr params.
-    # config.solr_path = 'select'
+    config.solr_path = 'search'
```
cdmo, line 67: This is how this setting really ought to have always been. We use a custom handler found at /search, we should specify it here. I think Solr was working because in our solr config we specify in the `search` handler that it is default, but this to me is clearer, and less working-by-accident. Haven't look at this in a while though so please let me know if I am forgetting/missing something!

app/views/bento/index.json.jbuilder
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
```
cdmo, line 1: I started with [the json builder in blacklight](https://github.com/projectblacklight/blacklight/blob/master/app/views/catalog/index.json.jbuilder) and trimmed what we don't need.

config/routes.rb
```diff
@@ -19,6 +19,10 @@
     concerns :range_searchable
   end
 
+  resource :bento, controller: :bento, defaults: { format: :json }, only: [:index], as: 'bento', path: '/bento' do
```
cdmo, line 22: specify `controller` so it can be singular and not `BentosController`

solr/conf/solrconfig.xml
```diff
@@ -76,7 +76,7 @@
   </admin>
 
 
-  <requestHandler name="search" class="solr.SearchHandler" default="true">
+  <requestHandler name="/search" class="solr.SearchHandler" default="true">
```
cdmo, line 79: think we just must have typo'd not including the leading slash? It's needed

cdmo:
this is required for  https://git.psu.edu/i-tech/ul/-/issues/681

cdmo:
I liked using the controller spec for jbuilder as opposed to a view spec simply for ease of use really. It makes the controller spec an integration spec I guess, but it's just a lot simpler to pull of. No mocking of this and that needed. This [article is good](https://ahimmelstoss.github.io/blog/2014/07/27/testing-a-rails-api-with-rspec/) and [this one too](https://medium.com/@lcriswell/rails-api-request-specs-with-rspec-effeac468c4e).

config/routes.rb
banukutlu: makes sense!

cdmo:
This is being consumed by a feature branch of `ul` which is live on ul01test.libraries.psu.edu

app/controllers/bento_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+      format.atom { render status: :unsupported_media_type }
+      format.json do
+        @presenter = Blacklight::JsonPresenter.new(@response,
+                                                   blacklight_config)
+      end
+    end
+  end
+
+  blacklight_config.configure do |config|
+    config.add_facet_fields_to_solr_request = false
```
banukutlu, line 20: awesome short config :)

app/controllers/catalog_controller.rb
banukutlu: yes, since `/search` handler has default set up to it, it was working but having it explicitly set here is nice and clear, I agree 👍

solr/conf/solrconfig.xml
banukutlu: good catch, idk how this was OK without the `/` 🤔

cdmo: crazy right? I think the whole `default` thing really saved us. When you do a search and tail solr, you'll see the `select` handler is used, but, this `search` one is actually used. It is bizarre and "working by accident" I think. 🙃

banukutlu: woww 😬

spec/controllers/bento_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      it 'has the expected keys' do
+        expect(json_response.keys).to match(['continueSearch', 'meta', 'data'])
+      end
+
+      it 'has a continue search link' do
+        expect(json_response['continueSearch']).to include('http://test.host/catalog?q=green&utf8=%E2%9C%93')
+      end
+
+      it 'has a meta key with metadata containing total count of results' do
+        expect(json_response['meta']&.dig('pages', 'total_count')).to be(7)
```
banukutlu, line 36: the number of results is dependent on our sample data we use as fixtures, right?

cdmo: yes

solr/conf/solrconfig.xml
cdmo: :nods:

Issue #585, opened 2020-06-03
HathiTrust links


cdmo:
Requires https://github.com/psu-libraries/psulib_traject/issues/204 first

Per record, create a link direct to HathiTrust catalog record or scanned item with appropriate text accompanying

Use https://github.com/psu-libraries/psulib_blacklight/wiki/Google-Books-and-HathiTrust-Integration-Links#enhancement-splitting-hathitrust-etas-and-public-domain for requirements

## Include on/off toggle in `Settings`

When on: etas links for denies
When off: denies become "search inside" links at the record view and nothing shows for denies at the index view

## Link logic

```
allow + mono = link to scan with Access digital copy through HathiTrust
allow + multi = link to catalog record page with Access digital copy through HathiTrust
deny + mono = link to scan with ETAS
deny + multi = link to catalog record in ht with ETAS
```

cdmo:
@ruthtillman I think I'm not yet clear on what exactly we're trying to do with the records in the overlap report that have the value "allow" in the "access" field (i.e., our "full view forever" items). I've looked at a couple records in the CAT that show up in the overlap report, for example:

*The township geology of Elk and Forest counties*...
https://cat.libraries.psu.edu/uhtbin/cgisirsi/x/0/0/57/5/0?searchdata1=1000103%7BCKEY%7D&searchfield1=GENERAL%5ESUBJECT%5EGENERAL%5E%5E&user_id=WEBSERVER

This item is in the overlap report as access:allow. It is also in HT: https://catalog.hathitrust.org/Record/007043464?type%5B%5D=all&lookfor%5B%5D=318366827&ft= (you can see the scan came from us).

Our record view in the CAT there is only a "Check out digital copy through HathiTrust" link that goes straight to the Hathi reader for that item. That language, to me, doesn't make a ton of sense seeing as how you don't need to check the book out, you can just read it straight away. Did we land on how we wanted to treat this link in the catalog? I think at this point given that we are even more explicitly telling patrons that the link is something they are checking out because of ETAS in the catalog, and because we've already got the data at the ready, I'm thinking that it'd be better user experience to just create a different treatment for full-text-always links. Something like "[HT logo] Access online throught HathiTrust."

[...]

PR #595, opened 2020-06-24
585 hathi links

banukutlu:
Issue #585


banukutlu:
@cdmo Would you mind doing a quick review for this? It is still WIP but I'd like to get your thoughts/feedback before moving on with the rest.

cdmo:
Sure thing

cdmo:
Or I guess you're not planning on adding the Solr field `hathi_struct` to the catalog controller? I think I was expecting something similar to the way we process other fields in Solr, and `full_links_struct` for example.

app/controllers/catalog_controller.rb
```diff
@@ -181,6 +181,7 @@ def index
     config.add_index_field 'publication_display_ssm', label: 'Publication Statement', sr_only: true
     config.add_index_field 'edition_display_ssm', label: 'Edition', sr_only: true
     config.add_index_field 'full_links_struct', label: 'Access Online', helper_method: :generic_link
+    config.add_index_field 'hathitrust_struct', label: 'HathiTrust Access', helper_method: :hathi_link, hide_label: true
```
cdmo, line 184: Can you use the `sr_only` flag instead? I replaced the "hide label" functionality in this #580

banukutlu: just tried with `sr_only` and it did not hide the label, is there sth else to do with that flag to hide the label?

cdmo: I see, ok. Well it should be working fine for the index view at least. The sr_only logic would have to be added to the show partials too.

Leaving the label out creates an accessibility problem and a semantic web problem. You'd have a `dl` child item with only a `dd` and no `dt` in cases of "hiding the label."

banukutlu: ahh ok, I will add that logic to show as well then use it.

app/views/catalog/_show_default.html.erb
```diff
@@ -2,7 +2,7 @@
 <%# default partial to display solr document fields in catalog show view -%>
 <dl>
   <% doc_presenter.fields_to_render.each do |field_name, field| -%>
-    <dt class="blacklight-<%= field_name.parameterize %>-label"><%= render_document_show_field_label document, field: field_name %></dt>
-    <dd class="blacklight-<%= field_name.parameterize %> ml-5"><%= doc_presenter.field_value field %></dd>
+      <dt class="blacklight-<%= field_name.parameterize %>-label"><%= render_document_show_field_label document, field: field_name %></dt>
+      <dd class="blacklight-<%= field_name.parameterize %> ml-5"><%= doc_presenter.field_value field %></dd>
```
banukutlu, line 6:
```suggestion
    <dt class="blacklight-<%= field_name.parameterize %>-label"><%= render_document_show_field_label document, field: field_name %></dt>
    <dd class="blacklight-<%= field_name.parameterize %> ml-5"><%= doc_presenter.field_value field %></dd>
```

solr/conf/solrconfig.xml
```diff
@@ -224,7 +224,8 @@
          full_links_struct,
          isbn_valid_ssm,
          oclc_number_ssim,
-         lccn_ssim
+         lccn_ssim,
+         hathitrust_struct,
```
banukutlu, line 228:
```suggestion
         hathitrust_struct
```

app/views/catalog/_show_default.html.erb
cdmo: :)

cdmo:
Fleshed out `ht_link_text` as per https://github.com/psu-libraries/psulib_blacklight/pull/597#issuecomment-651769758 in f1ab9e1

app/views/catalog/_show_top_fields.html.erb
```diff
@@ -3,8 +3,14 @@
...
   <% blacklight_config.show_fields.each do |field_name, field| -%>
     <% if field.top_field && !document[field_name].nil? %>
-        <dt class="blacklight-<%= field_name.parameterize %>-label col-md-3"><%= render_document_show_field_label document, field: field_name %></dt>
-        <dd class="blacklight-<%= field_name.parameterize %> col-md-9"><%= doc_presenter.field_value field %></dd>
+      <dt class="blacklight-<%= field_name.parameterize %>-label col-md-3
+          <%= 'text-muted' unless field[:sr_only] -%>
+          <%= 'sr-only' if field[:sr_only] -%>">
+        <%= render_document_show_field_label document, field: field_name %>
+      </dt>
+      <dd class="blacklight-<%= field_name.parameterize %> col-md-9"><%= doc_presenter.field_value field %></dd>
```
cdmo, line 11: We don't have any `sr_only` fields as show fields, but, if we do in the future we'll have this all set for it.

PR #597, opened 2020-06-30
Removes the hathi_links struct from the fields list and instead just …

cdmo:
…renders the hathi_links method that's been included on the SolrDocument

What do you think of this instead? I kinda like it because it keeps all the markup in views and it limits the exposed (i.e. not `private`) methods available - also makes testing easier I think. Just kinda sends the document what it needs.

This is incomplete, I only got it working with one combo at the index view, but the rest should be pretty easy. And the specs.

#595

### DIFF
--- a/app/models/hathi_links.rb
+++ b/app/models/hathi_links.rb
@@ -1,54 +1,67 @@
 # frozen_string_literal: true
 
-class HathiLinks
-  attr_reader :document
+module HathiLinks
+  def hathi_links
+    return nil unless has_hathi_links?
 
-  def initialize(document)
-    @document = document
-    @ht_hash = ht_hash
+    {
+      text: ht_link_text,
+      url: ht_url,
+      additional_text: ht_additional_text
+    }
   end
 
-  def ht_hash
-    JSON.parse(document['hathitrust_struct']&.first)
-  end
+  private
 
-  def id
-    ht_hash['ht_id']
-  end
+    def has_hathi_links?
+      @_source['hathitrust_struct'].present?
+    end
 
-  def bib_key
-    ht_hash['ht_bib_key']
-  end
+    def hathi_links_field
+      @_source['hathitrust_struct']
+    end
 
-  def access
-    ht_hash['access']
-  end
+    def ht_hash
+      JSON.parse(hathi_links_field&.first)
+    end
 
-  def url
-    url = if id.present?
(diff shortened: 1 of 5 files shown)


app/models/hathi_links.rb
```diff
@@ -1,54 +1,67 @@
...
+    end
 
-  def etas_item?
-    access == 'deny' && Settings&.hathi_etas
-  end
+    def ht_link_text
+      I18n.t('blackcat.hathitrust.public_domain_text') if ht_access == 'allow'
+      I18n.t('blackcat.hathitrust.restricted_access_text') if ht_access == 'deny' && !Settings&.hathi_etas
+
+      I18n.t('blackcat.hathitrust.etas_text')
```
banukutlu, line 55: would not this always return etas_text? 🤔

banukutlu:
Sure, it looks good to me, my initial approach was similar in terms of not adding the fields to controller and having a partial instead, I personally liked this better since with this there is no need for helper methods and feels cleaner. Thanks for the refactoring the class as well, I was considering making it a module as well, looks better right like this 👍

app/models/hathi_links.rb
cdmo: Yep. Did that one too quick

cdmo:
Cool. Yeah treating as a field is just too cumbersome because it’s so specialized and customized. I needed to play with the code a bit to see myself I think. Glad you like the module approach better. I definitely like using the presenter approach over helper in general but also because it’s the pattern core blacklight espouses. 

So how to proceed. Want me to finish fleshing this out? You take it? We could split it?

app/models/hathi_links.rb
banukutlu, line 55:
```suggestion
      I18n.t('blackcat.hathitrust.etas_text') if etas_item?
```

banukutlu:
> customized. I needed to play with the code a bit to see myself I think. Glad you like the module approach better. I definitely like using the presenter approach over helper in general but also because it’s the pattern core blacklight espouses.
> 
> So how to proceed. Want me to finish fleshing this out? You take it? We could split it?

I can take it, I had already started working on the specs as well - will make changes to them after this refactor as well. I won't be much working today so will work on it tomorrow :)

cdmo:
This is a PR to be merged into your feature branch, so it probably doesn't need fixes before merge - you can just update locally, right? I'm just gonna merge it over to your feature branch and close this out.

PR #598, opened 2020-07-09
Gem update

cdmo:


cdmo:
erb-lint was 

```
If this file cannot be processed by erb-lint, you can exclude it in your configuration file.
undefined method `reject' for nil:NilClass
```

I dont have the output of the error on the reason I exclude the tag rule in rubo, but it was along the lines of ActionView wrong number of arguments.

PR #610, opened 2020-08-10
added link to legacy catalog

ruthtillman:
Per discussion

### DIFF
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -3,6 +3,7 @@
   <%= link_to 'check out our discovery roadmap', 'https://libraries.psu.edu/library-search-roadmap' %>.</p>
 <p>You can also:</p>
 <ul>
+  <li><%= link_to 'Search the legacy CAT', 'https://cat.libraries.psu.edu' %></li>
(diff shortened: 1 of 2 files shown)


app/views/catalog/_home_text.html.erb
```diff
@@ -3,6 +3,7 @@
   <%= link_to 'check out our discovery roadmap', 'https://libraries.psu.edu/library-search-roadmap' %>.</p>
 <p>You can also:</p>
 <ul>
+  <li><%= link_to 'Search the legacy CAT', 'https://cat.libraries.psu.edu' %></li>
```
cdmo, line 6: I wonder if it would be better to just say "Search The CAT (legacy system)" (or something) just to avoid folks calling the new catalog "the new cat"? Thinking out loud really.

"Search The CAT (legacy)"?
"Search The CAT (legacy catalog)"?  
Or just "Search The CAT"?

🤔

banukutlu: I personally liked "Search the CAT (legacy catalog)"

ruthtillman: I tried making that change but it needs review bloew?

Issue #611, opened 2020-08-10
Previous/next session based 


cdmo:
From INC1493779

Parent issue https://github.com/projectblacklight/blacklight/issues/2295

cdmo:
@ruthtillman this is ready for review on QA. Do a search with some number of results, click a result, copy the URL, open new tab and paste it in: the previous/next links should not be there. In the production catalog they would be.

ruthtillman:
Testing:
1. Original container: If I am in the Firefox container in which the search was performed, no matter how many new tabs I open and paste in the URL, I still DO see prev/next. Those prev/next do work for that original search.
2. New container: If I go paste the URL in a new tab in a new container, I DO NOT see prev/next.
3. Original container after viewing unrelated record: If I go to a completely unrelated record in Blackcat in the original container by pasting in a known URL, I DO NOT see prev/next. If I THEN paste in the original URL after viewing the unrelated record, I DO NOT see prev/next.
4. Original container after viewing _related_ record: If I page through results then _paste_ in the URL of the result in which I'm interested (vs. paging back to it), I DO NOT see prev/next.

Assessment: This addresses the concern. Prev/Next are **not** persisting after you've looked at another record, only if you return to one you were looking at earlier without looking at any other record in the interim (and I would guess there's probably also an expiration on that, but I don't have a few days to let it go stale).

cdmo:
Yep, my scenario that I asked to be tested actually fails. What passes is now is that if you:

1. Do a search
1. Click on a result -> the page will have a prev/next link set
1. Go directly in a new tab _any_ other catalog record -> this won't have any prev/next toggle

In the catalog production right now:

1. Do a search
1. Click on a result -> the page will have a prev/next link set
1. Go directly in a new tab _any_ other catalog record -> this will have the same prev/next link as the item 2

So this problem is in _better_ shape for sure, but not fix entirely.

ruthtillman:
Yes, and it's not entirely a bug to put people back in their earlier session/search as long as we're not doing it completely randomly like it currently is. It's... an ambiguous feature 😅 . So I am ok to go ahead with this.

cdmo:
Yes the current craziness of the prev/next on stale sessions is almost completely fixed for sure. We'll deploy it up to prod and think about solving this last scenario another time. Thanks

PR #612, opened 2020-08-13
Adds document_id check as part of checking display of previous/next t…

cdmo:
…oggles

#611

Not ready for merge (why it's a Draft), but, wanted you to see this approach. It is working well for me locally.

Another thing to consider is that this could be fixed upstream first (though we'd have to upgrade a bunch)


cdmo:
We've upgraded a bunch now

cdmo:
@banukutlu can you take a look at this and let me know what you think?

It solves the problem of a direct link into the catalog (like from bento) showing incorrect prev/next link toggles. There's an upstream PR in core Blacklight that this is copying but I'm not sure that will get merged, it's just a draft.

app/controllers/catalog_controller.rb
```diff
@@ -45,6 +45,21 @@ def index
     end
   end
 
+  def track
```
banukutlu, line 48: overriding this method to add document_id 👍   I know you haven't tidy up the PR yet but just thought i'd mention it would be nice to add a comment to point that it is an override.

app/helpers/url_helper.rb
```diff
@@ -0,0 +1,19 @@
+module UrlHelper
```
banukutlu, line 1: same for this, comment would be nice

cdmo: Sure, for both. You feeling okay-ish about this PR in general? Shall I proceed? Or should we try something else? I did think briefly about attempting to make the prev/next toggles take on the context of the bento search click. But I wasn't sure that would even be desired or expected and given that I think it'd be a fair amount of work to implement (possibly?)

banukutlu: Yes, I think this approach makes sense and it is the upstream sol'n just not merged yet. Let's go with this, once it is merged, we will take out our overrides 👍  thanks!

banukutlu: it makes sense to not show a the prev/next when the user not coming from a search context within BL, i think it is OK to hide them when coming from bento

cdmo: cool! Thanks Banu

cdmo: > same for this, comment would be nice

We don't add a comment for the other "blacklight helpers" that we override - facets and layouts. I think we're ok without a comment for this one. ?

banukutlu (approved):
great soln, thanks @cdmo

Issue #615, opened 2020-08-15
Refactor production logging


cdmo:
Include hostname as prefix to log name

- [ ] Examine log level, requested that we lower it to reduce size

(h/t @whereismyjetpack)

PR #616, opened 2020-08-17
Ratchets production log to :warn and "paramaterizes" the production l…

cdmo:
…og output file to be named for the current hostname.

#615 

Note: we lose visability on the time it takes for things to render, which is a little concerning to me, but the logs are apparently just too expensive (40GB a day!). Hopefully we'll have datadog APM in place soon and then we won't miss that render time details


whereismyjetpack:
Is the render time the view render time?

cdmo:
Yes. And that graph looks 🔥 

Looking forward to that contract finalizing or whatever. Thanks.

banukutlu:
Also leaving this here https://github.com/roidrage/lograge just to note

whereismyjetpack:
In ScholarSphere, we are using lograge, with a custom JSON formatter to emit logs into a machine readable format. 

https://github.com/psu-stewardship/scholarsphere-4/pull/240

cdmo:
Ok. I'm proposing we use `lograge` and set level to `:info`. `:info` gives us far less verbose logs *but still gives us some insight into the performance of requests*. Basically each http request on the application gets an executive summary.

`:debug` in lograge, to me, isn't great, so I don't want to use it locally. Rails default logger gives better details, specifically the render time of each discrete partial.

whereismyjetpack:
For local development we flip lograge off to get the goods locally. LGTM

config/environments/production.rb
```diff
@@ -81,13 +81,26 @@
...
     logger.formatter = config.log_formatter
     config.logger    = ActiveSupport::TaggedLogging.new(logger)
   end
 
+  # Adds lograge and parametrizes log name. Capture parameters other than `controller` and `action` in lograge. ALso
+  # captures timestamp.
+  config.lograge.enabled = true
+  config.lograge.custom_options = lambda do |event|
+    params = event.payload[:params].reject { |k| %w(controller action).include?(k) }
+    {
```
banukutlu, line 95: thanks @cdmo!

Issue #505, opened 2019-10-31
My Account integration with I want It


banukutlu:
pending https://github.com/psu-libraries/myaccount/issues/96

* AJAX or loaded request to my account (loaded being the referrer parameter)
* Make blackcat qa talk to myaccount qa, could do a config based/modular approach (pick your my account server kinda thing). But this is to enable complete end to end testing of the qa environment of blackcat and myaccount.

Aha! Link: https://penn-state-university-libraries.aha.io/features/STPROJ-18

cdmo:
Can we close this Banu? It appears to be working and there's nothing for Ruth to review on this one.

PR #618, opened 2020-08-18
Myaccount integration

banukutlu:
Issue #505

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,7 @@ coverage
 
 # Ignore config gem settings.yml, the application's configuration should not be tracked.
 config/settings.yml
+
+config/settings.local.yml
+config/settings/*.local.yml
+config/environments/*.local.yml
--- a/.gitignore
+++ b/.gitignore
@@ -42,9 +42,7 @@ spec/examples.txt
 # testing output
 coverage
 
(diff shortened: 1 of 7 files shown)


cdmo:
Looks good, I think I'd just put the prod url in settings.yml though and remove the production and development files entirely.

banukutlu:
> Looks good, I think I'd just put the prod url in settings.yml though and remove the production and development files entirely.

this is the part, I want to touch base with you actually, the reason why this is still draft. I added those mostly because I was trying to find a way to make sure we do not mistakenly use myaccount prod on our locals. We can override it with a local setting file but still can be missed 🤔   
Also blackcatqa would be using myaccountqa, I guess the deployment script will handle that setting? your thoughts?

cdmo:
I guess I don't feel strong one way or the other. I generally try to avoid specific environment settings so I'd plan to handle this with the ignored settings files locally. We can talk more over zoom orsomething

banukutlu:
> I guess I don't feel strong one way or the other. I generally try to avoid specific environment settings so I'd plan to handle this with the ignored settings files locally. We can talk more over zoom orsomething

sure, let's chat when we can

Issue #619, opened 2020-08-19
Do Not Display Availability section if Record has ETAS (Access: Deny)


ruthtillman:
Multiple people doing chat reference report patrons being unsure why they see "I Want It" for some items and not for others. Because items in ETAS are not available, even if they're in a location being paged, let's suppress their Availability display entirely instead of just suppressing the "I Want It" button. Same logic applies. The HathiTrust link will still display.

Notes:

1. We will not worry about item faceting based on location. This is a temporary fix for a pandemic-induced problem.
2. If an item is NOT in ETAS and NOT being paged (e.g. at a campus not doing curbside, etc.) we will not worry about it _for now_.

banukutlu:
pls refer to #629 for to review this feature, because the requirements changed slightly and details are in that new issue.

PR #620, opened 2020-08-19
hide availability for hathi etas items

banukutlu:
Issue #619

### DIFF
--- a/spec/features/availability_spec.rb
+++ b/spec/features/availability_spec.rb
@@ -52,8 +52,34 @@ RSpec.feature 'Availability', type: :feature do
 
     it 'that is an online resource and has no holdings to display' do
       visit '/?utf8=✓&search_field=all_fields&q=D-humanos+Arruti%2C+Mariana'
-      expect(page).not_to have_selector('button[data-target="#availability-22091400"]')
+      expect(page).not_to have_selector 'button[data-target="#availability-22091400"]'
       expect(page).not_to have_selector '.availability-22091400'
+    end
+
+    context 'when Hathi ETAS is enabled' do
+      before do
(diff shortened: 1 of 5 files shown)


spec/features/availability_spec.rb
```diff
@@ -52,9 +52,35 @@
...
     it 'that is an online resource and has no holdings to display' do
       visit '/?utf8=✓&search_field=all_fields&q=D-humanos+Arruti%2C+Mariana'
-      expect(page).not_to have_selector('button[data-target="#availability-22091400"]')
+      expect(page).not_to have_selector 'button[data-target="#availability-22091400"]'
       expect(page).not_to have_selector '.availability-22091400'
     end
+
+    context 'when Hathi ETAS is enabled' do
+      before do
+        Settings.readonly_holds = false
```
cdmo, line 61: I think I'd like to see this setting name change from to something like `Settings.display_holdings`.

banukutlu: I can change this but just to make it clear, not using this setting for this etas work.

banukutlu:
> I think this is fine and works, I would approve. I am ]starting to feel like we might want to collapse some of this if, if, unless stuff into either a helper(s) or maybe a view component?

yeah a view component would be nice, I will add an issue for a refactor, sounds OK?

spec/features/availability_spec.rb
cdmo: No I'm clear on that, just something that has been bugging me a bit, like always is hard to comprehend. Like what does it mean again? I always have to figure out again. Does it mean whether or not we display the button or the holds or? Just think it could be clearer.

Issue #632, opened 2020-08-24
Update CHECKEDOUT location description and link to ILLiad


banukutlu:
> ~RESERVE-EM and RESERVE-EG both have the following description - Unavailable for checkout, request pages via ILLiad~ (see #636 )
> 
> And CHECKEDOUT has a new description - Unavailable for checkout, request through Interlibrary Loan
> 
> Can we have ILLiad and Interlibrary Loan both link in the cat/catalog to Illiad?
> 
> Thanks!
> 
> Chris Holobar

banukutlu:
some examples for testing:

```
RESERVE-EG - catkey 1056505
RESERVE-EM - catkey 78798
CHECKEDOUT - see catkey 261, 107943, 108540
```

banukutlu:
URL for EG/EM reserve scans `https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll?Action=10&Form=30&Genre=GenericRequestReserves`

banukutlu:
>Just want to confirm the final language and linking for these locations:
>
>RESERVE-EM and RESERVE-EG both have the following description - Unavailable for checkout, request a chapter (link to ILLiad scan/chapter request form)
>
>CHECKEDOUT has a new description - Checked out, request through Interlibrary Loan (link to ILLiad book/loan request form)
>
>Thanks!
>
>Chris Holobar

ruthtillman:
QAd. Looks good.

PR #633, opened 2020-08-24
updating checked out label and adding it to request_via_ill list

banukutlu:
Issue #632

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: xenial
 language: ruby
 rvm:
-  - 2.5.3
+  - 2.6.6
 bundler_args: --without development
 sudo: false
 cache:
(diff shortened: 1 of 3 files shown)


banukutlu:
waiting on the new ILL form url for RESERVE-EG and RESERVE-EM locations

cdmo:
I just now noticed that travis is on our old version of ruby, so updated. Build was failing anyway so don't think you would mind.

banukutlu:
> I just now noticed that travis is on our old version of ruby, so updated. Build was failing anyway so don't think you would mind.

sure of course not 👌

cdmo:
I checked the [3 listed here](https://github.com/psu-libraries/psulib_blacklight/issues/632#issuecomment-679334721) and they are not in the local index. Are there examples of this locally?

banukutlu:
> I checked the [3 listed here](https://github.com/psu-libraries/psulib_blacklight/issues/632#issuecomment-679334721) and they are not in the local index. Are there examples of this locally?

yes, examples from our local sample `22080733` and `24053587`

PR #638, opened 2020-08-26
Mitigates serialize-javascript CVE

cdmo:
* Loosens restrictions on webpacker yarn and gem packages and updates both

Also removed `webpack` as a discrete package. `webpack` gets included by `@rails/webpacker`.


cdmo:
I ended up needing to upgrade our version of node (might as well anyway I suppose). You'll need to run `nvm install` locally. When we go to deploy we'll need to make sure the deploy script refactor can handle this change too.

This was the problem

```
➜  psulib_blacklight git:(serialize-js-bump) yarn
yarn install v1.21.1
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error browserslist@4.14.0: The engine "node" is incompatible with this module. Expected version "^6 || ^7 || ^8 || ^9 || ^10 || ^11 || ^12 || >=13.7". Got "13.6.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

Issue #636, opened 2020-08-25
Update RESERVE-EM and RESERVE-EG location descriptions and link to Illiad


ruthtillman:
See #632 for context. We are awaiting link text and URL confirmations from the ILL department, who need to explore them further.

sample records:
```
RESERVE-EG - catkey 1056505
RESERVE-EM - catkey 78798
````

Edit:

```
RESERVE-EM and RESERVE-EG both have the following description - Unavailable for checkout, request a chapter (link to ILLiad scan/chapter request form) 

https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/OpenURL?Action=10&Form=30&Genre=GenericRequestReserves

CHECKEDOUT has a new description - Checked out, request through Interlibrary Loan (link to ILLiad book/loan request form)

https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/OpenURL?Action=10&Form=30
```

ruthtillman:
Glad we got the language. QAd and looks good!

PR #639, opened 2020-08-26
adding two reserve location to ILL list, also new ILL form for them

banukutlu:
Issue #636

### DIFF
--- a/spec/features/availability_spec.rb
+++ b/spec/features/availability_spec.rb
@@ -321,6 +321,9 @@ RSpec.feature 'Availability', type: :feature do
         end
       end
 
+      it 'has the correct ILL url if a record has a holding in a CHECKEDOUT location'
+      it 'has the correct ILL url if a record has a holding in a reserves scan location'
+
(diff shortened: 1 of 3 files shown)


spec/features/availability_spec.rb
```diff
@@ -321,6 +321,9 @@
         end
       end
 
+      it 'has the correct ILL url if a record has a holding in a CHECKEDOUT location'
+      it 'has the correct ILL url if a record has a holding in a reserves scan location'
```
cdmo, line 325: These are stubbed because we want to rush this change and get the specs in later?

cdmo:
need to merge #640 first

spec/features/availability_spec.rb
banukutlu: I was also thinking adding after we refactor these specs to use Sinatra, because you know items changing locations causing specs to fail right now.

cdmo: perfect sounds good

PR #643, opened 2020-08-31
Instructs webpacker to compile changes made to .json files

cdmo:
* Also prunes unneeded rake task


cdmo:
This needs to go in so that changes in last PR actually trigger a compile

Issue #644, opened 2020-08-31
Add News & Microforms Scan Request


ruthtillman:
Add News & Microforms Scan Request to ILL/Aeon linking logic. See Wiki https://github.com/psu-libraries/psulib_blacklight/wiki/News-Microforms-Scan-Request

## Conditions

IF Item Library == UP-MICRO
AND Item Type == MICROFORM
AND Item Home location != THESIS-NML

then add the link below the call number display (same as Aeon links or Request Scan links)

## Link Values

Link text: `Request Scan - Penn State Users`

Link: `https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll?Action=10&Form=30&Genre=GenericRequestMicroScan&title=`title_display_ssm`&callno=`{holding.callNumber}`loc=2nd%20Floor%20Paterno%20-%20Microfilm&rfr_id=info%3Asid%2Fcatalog.libraries.psu.edu`

Example link: 

https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll?Action=10&Form=30&Genre=GenericRequestMicroScan&title=General%20office%20file.%20%5BStaff--Roy%20Wilkins%5D%20%5Bmicroform%5D&callno=E185.5.N276A2%20pt.30%201966-72%20ser.A%20r.18&loc=2nd%20Floor%20Paterno%20-%20Microfilm&date=%5B2002%5D&rfr_id=info%3Asid%2Fcat.libraries.psu.edu

cdmo:
Deployed to QA, seems to be working https://blackcat01qa.libraries.psu.edu/catalog/13855626. @ruthtillman

ruthtillman:
Looks great! I went and tested a) how it appears in search results 👍 and b) a thesis to make sure they still look right. :shipit:

Issue #631, opened 2020-08-24
Sinatra for web service responses in feature specs


cdmo:
Would like us to do this so we don't have to play whack a mole with flakey specs that change depending on the status of the real live holding data

cdmo:
Kinda curious - are you also moving the availability js work into a rails controller? I am not sure sinatra would work unless we did that because the web service requests come from the browser right? Just wondering the size of this issue - it might be big.

nicktiberi:
@banukutlu - I think we can close this since we are using VCR to mock the availability responses. Do you agree?

PR #650, opened 2020-09-02
adds news and microforms ILL link

banukutlu:
Issue #644

### DIFF
--- a/app/javascript/availability/index.js
+++ b/app/javascript/availability/index.js
@@ -502,7 +502,7 @@ const availability = {
                     if (linkType === "reserves-scan") {
                         ILLURL += `&Genre=GenericRequestReserves&location=${itemLocation}`;
                     }
-                    if (linkType === "news-microform") {
+                    if (linkType === "news-microform-scan") {
                         ILLURL += `&Genre=GenericRequestMicroScan&location=${itemLocation}`;
                     }
                     ILLURL += `&title=${title}&callno=${callNumber}&rfr_id=info%3Asid%2Fcatalog.libraries.psu.edu`;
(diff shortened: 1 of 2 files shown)


spec/features/availability_spec.rb
```diff
@@ -329,5 +329,9 @@
         expect(page).not_to have_link 'Copy unavailable, request via Interlibrary Loan'
       end
     end
+
+    describe 'News & Microforms Scan Request' do
+      it 'display the correct ILL url'
```
cdmo, line 334: Why a pending spec here again? Because you plan to write the sinatra fake?

banukutlu: yes, I didn't want to add new fixtures to our sample extract just for these cases. I would need to ask for a new extract for our sample with catkeys for these cases, and then they might change even 🤔

cdmo: ok, marking this as part of #631

app/javascript/availability/index.js
```diff
@@ -406,7 +406,7 @@ const availability = {
 
     generateCallNumber(holding) {
         // Do not display call number for on loan items
-        return (holding.locationID === 'ILLEND') ? "" : holding.callNumber;
+        return (holding.locationID === "ILLEND") ? "" : holding.callNumber;
```
cdmo, line 409: was this a lint thing? single versus double quotes?

banukutlu: ohhh this was just me trying to attempt for consistency but then I see that there are single and double in multiple places 😬

cdmo (approved):

PR #667, opened 2020-09-28
Adds scan_sim, the special collections accessioning number, to search…

cdmo:
… algorithm

Requires https://github.com/psu-libraries/psulib_traject/pull/247

Not that this adds `scan_sim` to the Identifiers Advanced Search field too

### DIFF
--- a/solr/conf/solrconfig.xml
+++ b/solr/conf/solrconfig.xml
@@ -119,6 +119,7 @@
          series_title_tsim^10
          isbn_sim
          issn_sim
+         scan_sim
          toc_ssim
          notes_summary_ssim
          all_text_timv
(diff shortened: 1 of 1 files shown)


cdmo:
Once psu-libraries/psulib_traject#247 is merged and deployed to QA, and once QA has been reindexed to include this new field, I'd like to deploy this feature branch to QA and do some tests on a full index _before merging_

banukutlu:
@cdmo This looks good to me but do you wanna test it on QA before marking it for review?

cdmo:
I guess it's ok to merge 🤷 

I think I was just thinking it's something that's hard to test and do dev on without a full nearly 8 million record index to work with. But does seem to me that this is fine the way it is. If you're good with it I am.

banukutlu (approved):
I think it looks good 👍

Issue #666, opened 2020-09-25
Document how sample marc data is updated, how fixtures are updated. Also, generate fixtures appears to be not working


cdmo:
```
➜  psulib_blacklight git:(master) ✗ bundle exec blackcat:traject:generate_fixtures
bundler: command not found: blackcat:traject:generate_fixtures
Install missing gem executables with `bundle install`
```

not sure what's up here yet, just recording the problem

We need some docs I think too, I am not finding anything.

banukutlu:
I think this is the cmd `bundle exec rake blackcat:traject:create_fixtures`, it worked for me 🤔

cdmo:
Ah dumb me I was using the wrong command! I was using a command in my bash history I found, don't know why it was there!

Anyway, using the right command,  now I am getting the following:

```
➜  psulib_blacklight git:(scan_sim-addition) bundle exec rails blackcat:traject:create_fixtures
[DEPRECATED] `Bundler.with_clean_env` has been deprecated in favor of `Bundler.with_unbundled_env`. If you instead want the environment before bundler was originally loaded, use `Bundler.with_original_env` (called at /Users/cdm32/Projects/ruby/psulib_blacklight/lib/tasks/blackcat.rake:33)
Could not find chronic-0.10.2 in any of the sources
Run `bundle install` to install missing gems.
```

The `psulib_traject` app is bundle up-to-date'd.

So, is it possible to add to `sample_psucat.mrc`? Or how does that get modified? Let's say I want to add a few OA access examples. How would I go about that? MarcEdit?

Also, why do we deindex at the beginning of the `create_fixtures` task? I think that can be removed, but maybe I'm not understanding what's going on. Seems like it has nothing to do with creating/generating fixtures? 🤔

banukutlu:
> Ah dumb me I was using the wrong command! I was using a command in my bash history I found, don't know why it was there!
> 

No worries, I think it was named liked that at some point and we changed.

> Anyway, using the right command, now I am getting the following:
> 
> ```
> ➜  psulib_blacklight git:(scan_sim-addition) bundle exec rails blackcat:traject:create_fixtures
> [DEPRECATED] `Bundler.with_clean_env` has been deprecated in favor of `Bundler.with_unbundled_env`. If you instead want the environment before bundler was originally loaded, use `Bundler.with_original_env` (called at /Users/cdm32/Projects/ruby/psulib_blacklight/lib/tasks/blackcat.rake:33)
[...]

Issue #665, opened 2020-09-22
Add display for fields 255, 775, and 793


ruthtillman:
See https://github.com/psu-libraries/psulib_traject/issues/244 for indexing.

Their locations for display are here: https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields with temporary notes dated 2020-09-22 to make them easy to find.

Note: link 793 display to a keyword quote search for the term. e.g. "Allison Shelley"

cdmo:
@ruthtillman regarding

> Note: link 793 display to a keyword quote search for the term. e.g. "Allison Shelley"

What are we expecting with the results? I assume that https://catalog.libraries.psu.edu/?search_field=all_fields&q=%22Allison+Shelley%22 is not good enough but I wanted to confirm. This search will find all sorts of matches including those records that have a collection matching "Allison Shelley" but any other record that matches "Allison Shelley" based on that phrase appearing elsewhere. Maybe this is ok? Or would we rather be more restrictive about the search (i.e., specifying that field only)

ruthtillman:
It should be ok. Allison Shelley is quite a large collection. Let's start with that and I'll test it to see if we need to refine -- unless it's easier to just do the more restrictive linking.

cdmo:
I made the collection a hidden facet, much like how we treat authors.

ruthtillman:
These all look good!

PR #668, opened 2020-09-30
665 editions

cdmo:
#665 

requires https://github.com/psu-libraries/psulib_traject/pull/250 

also #666


cdmo:
I wonder if we should ask Maryam for another update to our sample marc. The following collection facet data is not present in our current sample, I had to import it by just importing the beginning part of the last full extract.


Screen Shot 2020-09-30 at 9 57 35 AM

banukutlu (approved):
👍

Issue #522, opened 2019-11-26
Update Logic for I Want It & Include Item-Level Logic


ruthtillman:
[note: we need to verify some behavior re: whether an item currently On Reserve and Checked Out would display as holdable or not.]

## Item Level Logic - Multiple Call Numbers

When a record has more than one unique call number value (note, this is not the same as call number record, since the same call number record may belong to 3 different libraries), e.g. 

```
PN1997.A3229 2003 DVD disc.1-2
PN1997.A3229 2003 DVD disc.3-4
PN1997.A3229 2003 DVD disc.5-6
PN1997.A3229 2003 DVD disc.7-8
```

Patrons should be able to select each call number separately when making a request. They may select all or just one. This should be determined by the currentLocationID and homeLocationID:

* If homeLocationID and currentLocationID are [holdable locations](https://cat.libraries.psu.edu/uhtbin/Lists/locations), it should be displayed as a possible selection.
* If neither or only one is a holdable location, it should not be displayed as a possible selection.
* If a unique call number is in two locations, one of which is holdable and one of which is not is not, per logic above (e.g. one copy is in Reference at Berks and one is in General Stacks at Harrisburg), it should be allowed to be put on hold.

ruthtillman:
Wiki: https://github.com/psu-libraries/psulib_blacklight/wiki/Holds-and-ILL#missing-lost-and-other-unholdable-items

banukutlu:
@ruthtillman to clarity, which way do you prefer and would be more accurate `Checking whether it's on course reserve` or `Checking current and home locations`?

ruthtillman:
@banukutlu Let's do the option of: Checking whether it's on course reserve.

banukutlu:
@ruthtillman  re. Item level logic, isn't this handled in myaccount with https://github.com/psu-libraries/myaccount/issues/96?

ruthtillman:
This is an older ticket, so yes it looks like on the holds placement side we're set.

But we still need to not display "I Want It" if the available items are only on course reserve.

banukutlu:
@ruthtillman  do you have any course reserves catkeys?

banukutlu:
@mkutch could you help with finding/creating an example catkey, also additionally a bound-with example would be great as well.

banukutlu:
@ruthtillman how would this work for bound holdings? do we get the `reserveCollectionID` info from the parent?

ruthtillman:
that sounds right.

[...]

PR #670, opened 2020-10-02
do not display hold button if all items on course reserves

banukutlu:
#522


app/javascript/availability/index.js
```diff
@@ -610,6 +614,20 @@ const availability = {
 
     isArchivalMaterial(holding) {
         return (['UP-SPECCOL'].includes(holding.libraryID) && !availability.isMoved(holding.homeLocationID));
+    },
+
+    isHoldButton(holdings) {
```
cdmo, line 619:
```suggestion
    isHoldable(holdings) {
```

is that better?

or maybe `showHoldButton` (wish we could have question marks). what you have is probably fine

banukutlu: I like `showHoldButton` 👍

cdmo (approved):
👍

Issue #269, opened 2019-04-18
Advanced Search: Location


cdmo:
Born from #239
Requires https://github.com/psu-libraries/psulib_traject/issues/121

Aha! Link: https://penn-state-university-libraries.aha.io/features/STPROJ-48

ruthtillman:
Behavior expectations for Advanced Search by Location is here: https://github.com/psu-libraries/psulib_blacklight/wiki/Library-Faceting-and-Locations-Management#locations-advanced-search

ruthtillman:
Order of operations thought:

1. Index from 949l (L)
2. Exclude locations the excludes list
3. Map to our other locations label list.
4. Add as to advanced search.
5. Profit?

cdmo:
🤑

ruthtillman:
GORGEOUS!

PR #676, opened 2020-11-23
adds location to advanced search

banukutlu:
#269 

requires https://github.com/psu-libraries/psulib_traject/pull/264

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,7 +23,8 @@ Metrics/BlockLength:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 285
+  Exclude:
+    - 'app/controllers/catalog_controller.rb'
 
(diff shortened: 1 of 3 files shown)


.rubocop_todo.yml
```diff
@@ -23,7 +23,7 @@ Metrics/BlockLength:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 284
+  Max: 285
```
cdmo, line 26: wonder if we should add an exclude for catalog_controller.rb here too

cdmo (dismissed):
Looks good code-wise. Once the traject PR is merged I can give it a whirl locally.

.rubocop_todo.yml
banukutlu: yeah, makes sense, will do

cdmo (approved):

PR #684, opened 2020-12-11
Adds GitHub Action for CI and Solr management client

cdmo:


config/settings.yml
```diff
@@ -2,4 +2,4 @@ hold_button_url: http://cat.libraries.psu.edu/cgi-bin/catredirpg?C=
 matomo_id: 10
 solr:
   url: http://127.0.0.1:8983
-  collection: psul_blacklight
\ No newline at end of file
+  collection: psul_catalog
```
banukutlu, line 5: yup better name ✔️

cdmo: yeah figured I'd change this while I was at it

cdmo:
Another nice thing about this approach is that we don't have to worry about installing browsers, see https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#browsers-and-drivers

cdmo:
Added some docs https://github.com/psu-libraries/psulib_blacklight/wiki/Continuous-Integration

cdmo:
Testing of the SolrConfig app is done via being called in the SolrManager tests. I think this is okay because mostly it's just reading strings. But, there is the md5 hashing private method which may be worth testing, not sure.

cdmo:
Sorry, I think I need to expose a method for the cases where the catalog needs a config update and collection config update but not a fresh index.

cdmo:
I think we should add a co-written-by @whereismyjetpack on this when we merge it in. This is derivative of https://github.com/psu-stewardship/scholarsphere/blob/develop/lib/scholarsphere/solr_admin.rb. Thanks @whereismyjetpack

cdmo:
One thing I learned through testing is that the `modify_collection` method, which just updates the current collection's configset via the collections API, doesn't require a reload of the collection after the command runs.

cdmo:
https://github.com/psu-libraries/psulib_blacklight_deploy/pull/103 is making use of this PR

lib/psulib_blacklight/solr_config.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'zip'
+
+module PsulibBlacklight
+  class SolrConfig
+    CONFIG_PATH = '/solr/admin/configs'
+    COLLECTION_PATH = '/solr/admin/collections'
```
banukutlu, line 8: @cdmo I guess you keep them here as part of config but since these are only used by solr_manager what about moving these to solr_manager.rb?

cdmo: Semantics is the only reason there's another class really. SolrManager is the only thing that calls SolrConfig. Borrowing from the SS4 works, I think it's only two classes because it's too much for one class. So the more straight-config type stuff goes in SolrConfig, where the Manager is more the actions taken upon collections and configs. That's how I think about it. What do you think?

banukutlu: Yeah, I figured your thoughts on this and see your point, I am not picky about this, was just pointing it. What I am thinking is that these two won't get changed and would be a bit shorter when used in the manager class just like

`list = connection.get(CONFIG_PATH, action: 'LIST')`

lib/psulib_blacklight/solr_manager.rb
```diff
@@ -0,0 +1,142 @@
...
+      end
+
+      def collection_name_with_version
+        @collection_name_with_version ||= "#{config.collection_name}_v#{next_collection_version}"
+      end
+
+      def next_collection_version
+        return 1 if collections&.grep(/#{config.collection_name}/)&.empty?
+
+        collections.map { |version| version.scan(/\d/).first.to_i }.flatten.max + 1
```
banukutlu, line 70: @cdmo is collections here is sth like this ["psu_v10"]? 
When I tried ` ["psu_v10"].map { |version| version.scan(/\d/).first.to_i }.flatten.max + 1` it gives me 2, am I checking it wrong?

cdmo: I will leave now and never come back 🤦 🤦 🤦 🤦 byeeee

lib/psulib_blacklight/solr_manager.rb
```diff
@@ -67,7 +67,7 @@ def collection_name_with_version
       def next_collection_version
         return 1 if collections&.grep(/#{config.collection_name}/)&.empty?
 
-        collections.map { |version| version.scan(/\d/).first.to_i }.flatten.max + 1
+        collections.map { |version| version.scan(/\d+/).first.to_i }.flatten.max + 1
```
banukutlu, line 70: 👍

lib/psulib_blacklight/solr_manager.rb
banukutlu: 😛

banukutlu:
@cdmo is https://github.com/psu-libraries/psulib_blacklight/blob/a79b95f4ee9e1e6af81d32242d326b32980ecd6a/spec/lib/psulib_blacklight/solr_config_spec.rb supposed to be empty?

lib/tasks/docker.rake
```diff
@@ -57,15 +62,7 @@ namespace :docker do
...
-    print `docker run \
-            -d=true \
-            --name felix \
-            -it \
-            -e SOLR_HEAP=1G \
-            -p 8983:8983 \
-            -v "$(pwd)"/solr/conf:/myconfig \
-            solr:7.4.0 \
-            -DzkRun`
+    print `docker run -d -p 8983:8983 solr:7.4.0 -DzkRun`
```
banukutlu, line 65: @cdmo no need for `-e` anymore 👍

cdmo:
I was on the fence about testing the solr config class. That's probably why I left it in there. Since it mostly just reads things from the Config gem it seemed kinda pointless. But, wasn't totally sure.

lib/tasks/docker.rake
cdmo: And removed the mounting of the config because that's done via SolrManager now. Also removed the name on the instance.

banukutlu:
> I was on the fence about testing the solr config class. That's probably why I left it in there. Since it mostly just reads things from the Config gem it seemed kinda pointless. But, wasn't totally sure.

yes, it makes sense to me not to test it separately, it probably is covered by testing the manager 👍

lib/tasks/docker.rake
```diff
@@ -1,43 +1,48 @@
...
   end
 
-  task build: [:up_zk_config] do
-    Rake::Task['docker:create_collection'].invoke
+  task update_config: :environment do
+    solr_manager = PsulibBlacklight::SolrManager.new
+    solr_manager.modify_collection
+  end
+
+  task last_incremented_collection: :environment do
```
banukutlu, line 43: @cdmo where do we use this task?

spec/lib/psulib_blacklight/solr_manager_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+          .to_return({ body: File.read('spec/fixtures/solr/configset_list.json') })
+      end
+
+      it 'attempts to upload configset' do
+        expect(solr_manager.send(:config_sets)).to include('configset-eae0a4a2696affb9748c2d2b2ac44cdf')
+      end
+    end
+
+    context 'when solr is up and the current configset is present in solr' do
+      let(:spy_solr_manager) { instance_spy(described_class) }
```
banukutlu, line 42: nice 👍

spec/lib/psulib_blacklight/solr_manager_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+    end
+  end
+
+  describe '#create_alias' do
+    it 'creates OR updates an alias to the current collection with version name' do
+      expect(solr_manager.create_alias).to equal 200
+    end
+
+    context 'when there are not collections with prefixes matching our collection name scheme' do
```
banukutlu, line 59:
```suggestion
    context 'when there are no collections with prefixes matching our collection name scheme' do
```

banukutlu (commented):
only one question added, reviewed all, it looks great, thanks for the great work 👏

lib/tasks/docker.rake
cdmo: this is used in the "create collection and import to it" play in our ansible scripts repo. The idea being that the very latest solr collection is the one we always want to import to. I think that's right.

banukutlu: yes it makes sense thanks!

banukutlu (approved):
:clap:

PR #685, opened 2020-12-14
Fix HT URL suffix to separate query parameters

cdmo:


banukutlu:
https://github.com/psu-libraries/psulib_blacklight/blob/ht-suffix/spec/models/hathi_links_spec.rb#L71

needs `&` in the spec as well

banukutlu (changes requested):
see comment

banukutlu (approved):

PR #700, opened 2021-03-03
Update README.md

banukutlu:


cdmo (approved):
Heh, I don't even know if these instructions are accurate anymore. Guess you'll be walking Nick through it though and we'll find out

PR #701, opened 2021-03-03
fix to librarian view routing

banukutlu:
changed the routes a bit since marc_view was failing because of a mismatched constraint on id

### DIFF
--- a/spec/controllers/catalog_controller_spec.rb
+++ b/spec/controllers/catalog_controller_spec.rb
@@ -3,8 +3,6 @@
 require 'rails_helper'
 
 RSpec.describe CatalogController, type: :controller do
-  let(:doc_id) { '20049333' }
-
   describe 'index action' do
     it 'gets the homepage and renders only the homepage facets' do
       get :index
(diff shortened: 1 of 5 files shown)


cdmo:
I am not understanding the scenario where multiple punctuation marks would need to be addressed, did Ruth email you separately or something? I thought the problem was for stuff like "https://catalog.libraries.psu.edu/catalog/1220805." and "[https://catalog.libraries.psu.edu/catalog/1220805]" and "https://catalog.libraries.psu.edu/catalog/1220805,". 

Maybe we're worried about typos in LibGuides like "https://catalog.libraries.psu.edu/catalog/1220805],"?

banukutlu:
> I am not understanding the scenario where multiple punctuation marks would need to be addressed, did Ruth email you separately or something? I thought the problem was for stuff like "https://catalog.libraries.psu.edu/catalog/1220805." and "[https://catalog.libraries.psu.edu/catalog/1220805]" and "https://catalog.libraries.psu.edu/catalog/1220805,".
> 
> Maybe we're worried about typos in LibGuides like "[https://catalog.libraries.psu.edu/catalog/1220805]](https://catalog.libraries.psu.edu/catalog/1220805%5D),"?

Yes, this change is re to Ruth's feedback after her QA review, she mentioned the multiple punctuations, it is convo in Slack.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -28,11 +26,11 @@
 
   describe 'route check' do
     it 'routes /catalog/:id properly' do
-      expect(get: '/catalog/:id').to route_to(controller: 'catalog', action: 'show', id: ':id')
+      expect(get: '/catalog/id').to route_to(controller: 'catalog', action: 'show', id: 'id')
```
cdmo, line 29: did you just change this because it's weird to pass in the string value ':id' as an argument to the catalog controller? or was there something failing?

cdmo (approved):

spec/controllers/catalog_controller_spec.rb
banukutlu: for some reason that I could not figure out exactly why, it started to fail when I had the multiple punct changes. I kept it like this after reverting those changes because of what you mention like a bit weird string value like a symbol.

cdmo: ah of course, it saw the punctuation we were looking for and wigged out. makes sense! thanks. very odd that we ever passed in a symbol-like string here anyway 🤷 

thanks

PR #706, opened 2021-03-12
rails update to 6.0.3.5

banukutlu:
addresses vulnerabilities


config/application.rb
```diff
@@ -24,7 +24,7 @@
 module Psulcat
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults '6.0'
+    config.load_defaults 6.0
```
cdmo, line 27: Nice catch. Does this actually change behavior or is it just "more correct"?

banukutlu: I think it is correct either way, load_defaults does `to_s` to it anyway, I ran `bin/rails app:update` to see what might have been changed, just got this changed during that, thought it looks cleaner nothing else I think.

cdmo (approved):

PR #708, opened 2021-03-12
Availability: Show a message in the UI when an online item is on order

### LINKED ISSUE #704: Find way to display "On Order" for ONLINE items which are on order.
One clarification question we've had with the Catalog is what to do with items which are On-Order and also ONLINE. Nothing in the search result or item view page display indicates that they're not yet available. Because we don't display availability for ONLINE, it's not showing either the Availability or the link. This confuses users who think the item is broken somehow. It's a very reasonable concern.

**Logic:** If an item is ONLINE _and_ the ONLY currentLocationID on a record is ON-ORDER `<currentLocationID>ON-ORDER</currentLocationID>` (unique -- as there may be 5 of them but all ON-ORDER --shouldn't really be the case with ONLINE but let's unique it to be sure), then we display something indicating that it's On Order.

**Display options:** We could just display the normal Availability box and simply put a line in it which says "On Order." We could also display a box similar to the link boxes which says "Being acquired by the library"

**Sample record:**  https://catalog.libraries.psu.edu/catalog/33183518 (true on 2021-03-10)

Its webservices response:

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?><LookupTitleInfoResponse xmlns="http://schemas.sirsidynix.com/symws/standard"><TitleInfo><titleID>33183518</titleID><TitleAvailabilityInfo><totalCopiesAvailable>1</totalCopiesAvailable><libraryWithAvailableCopies>Online Resource</libraryWithAvailableCopies><totalResvCopiesAvailable>0</totalResvCopiesAvailable><locationOfFirstAvailableItem>ON-ORDER</locationOfFirstAvailableItem><holdable>true</holdable><bookable>false</bookable></TitleAvailabilityInfo><CallInfo><libraryID>ONLINE</libraryID><classificationID>AUTO</classificationID><callNumber>XX(33183518.1)</callNumber><numberOfCopies>1</numberOfCopies><ItemInfo><itemID>33183518-1001</itemID><itemTypeID>ONLINE</itemTypeID><currentLocationID>ON-ORDER</currentLocationID><homeLocationID>ON-ORDER</homeLocationID><chargeable>true</chargeable><fixedTimeBooking>false</fixedTimeBooking></ItemInfo></CallInfo><numberOfBoundwithLinks>0</numberOfBoundwithLinks></TitleInfo></LookupTitleInfoResponse>
```

nicktiberi:
Re: #704

When an item is online-only (i.e. NO physical copies available) AND all `currentLocationId` fields for that item are `ON-ORDER`, we now show a message in the UI to indicate that the item is being acquired by the libraries:

Screen Shot 2021-03-12 at 1 29 52 PM

### DIFF
--- a/app/javascript/availability/index.js
+++ b/app/javascript/availability/index.js
@@ -50,6 +50,7 @@ const availability = {
         if (titleIDs.length > 0) {
             let allHoldings = [];
             let boundHoldings = [];
+            let isOnlineOnOrderOnly = false;
             const sirsiRequestParams = titleIDs.map(url => '&titleID=' + url).join('');
 
             $.ajax({
@@ -77,6 +78,9 @@ const availability = {
                         if (numberOfBoundwithLinks > 0) {
                             boundHoldings = availability.getBoundHoldings(boundHoldings, titleInfo);
                         }
+
+                        // check to see if the item is only available online AND is in ON-ORDER status
+                        isOnlineOnOrderOnly = availability.getIsOnlineOnOrderOnly(titleInfo);
                     });
 
                     if (Object.keys(boundHoldings).length > 0) {
@@ -85,7 +89,7 @@ const availability = {
                     }
                     else {
                         // Print availability data
-                        availability.availabilityDisplay(allHoldings);
+                        availability.availabilityDisplay(allHoldings, isOnlineOnOrderOnly);
                     }
                 }, function() {
                     availability.displayErrorMsg();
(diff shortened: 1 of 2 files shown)


banukutlu:
@nicktiberi Would you pls rebase this branch to get the Github Actions fix from main so the checks would run for this PR? thanks!

banukutlu (changes requested):
@nicktiberi Would you mind trying to add a spec for this here https://github.com/psu-libraries/psulib_blacklight/blob/main/spec/features/availability_spec.rb?

banukutlu:
@nicktiberi It looks good to me overall, just one question for Ruth - https://github.com/psu-libraries/psulib_blacklight/issues/704#issuecomment-803759955

banukutlu (approved):
looks good, thank you @nicktiberi!

PR #712, opened 2021-03-24
Availability: Fix for constant loading spinners on search results page

nicktiberi:
Re: #704

More specifically target individual items in the availability logic so that we can accurately show which items are online + on-order only.

**Search Results View:**
Screen Shot 2021-03-24 at 11 25 25 AM

**Individual Item View:**
Screen Shot 2021-03-24 at 11 07 32 AM


banukutlu (approved):
looks great 👌  thanks @nicktiberi

PR #705, opened 2021-03-11
Improve "Access Online" Links UI

nicktiberi:
Re: #672

- Calls more attention to online resources by putting them in a callout box.
- Visually distinguishes resources in the Penn State Digital Collections from other external online resources.

**External Online Resource:**
Screen Shot 2021-03-11 at 9 11 58 AM

**Penn State Digital Collections Resource:**
Screen Shot 2021-03-11 at 9 22 22 AM

Issue #672, opened 2020-10-06
Review/revise link display to online content


ruthtillman:
Requires https://github.com/psu-libraries/psulib_traject/issues/277
- [x] Links to our digital collections should be cool and more prominent kind of like HathiTrust
- [x] Links to general online content should overall be more eye-catching and grab the user better.

ruthtillman:
1) Ruth, to-do create Traject ticket for the display label of 856.

2) Ruth, review other institutions' catalogs for online links.

ruthtillman:
We'll be indexing link labels and notes IF they exist. Their use should be as follows:

1. No link label, no link note: use MVP default of displaying the domain.
2. Link label, no link note: use link label.
3. No link label, link note: use MVP default of displaying the domain, followed by the link note.
4. Link label, link note: use link label as label for link, follow with link note.

ruthtillman:
(in progress)

## Display for Ordinary Links:

See comment above for the label/note aspects.

### Search Results 

Display "single record label" text within blue box a la (Online in these results) https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=Late+Soviet+Posters 

### Record display field

[let's look at the best way to display this? maybe just a slightly nicer version of this display? https://catalog.princeton.edu/catalog/11698380 or this https://searchworks.stanford.edu/view/13797514 ? ]

## Display for our Digital Collections:

- Blue box
- Box title: View in Penn State Digital Collections
- Lion shield to the left just like with our HathiTrust logo.

ruthtillman:
domain test -- so far have identifier:

- digital.libraries.psu.edu
- libraries.psu.edu/collections 

and am asking Kevin C.

ruthtillman:
Rather than using bullet points next to links in these boxes, let's use either https://fontawesome.com/icons/external-link-alt?style=solid or https://fontawesome.com/icons/link?style=solid (I am open to feedback on this)

ruthtillman:
For the links, we need to refine all the way to these domains for actual digital collections because we have other library website links that might not be digital collections.

```
digital.libraries.psu.edu
libraries.psu.edu/collections
```

This one is doing exactly what it should: https://blackcat01qa.libraries.psu.edu/catalog/119281, as is this https://blackcat01qa.libraries.psu.edu/catalog/8400751 

[...]

PR #714, opened 2021-03-25
Refactor external links

banukutlu:
refactoring for #705
Issue: #672

just a bit reorg for external links @nicktiberi could you pls review/check? what do you think?


spec/models/access_online_links_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-RSpec.describe AccessOnlineLinks do
+RSpec.describe ExternalLinks do
```
nicktiberi, line 5: Maybe we should rename this file to `external_links_spec.rb` to match the new name of the spec?

spec/models/access_online_links_spec.rb
```diff
@@ -2,7 +2,7 @@
...
   describe '#access_online_links' do
```
nicktiberi, line 6: Probably rename this as well to `#external_links`?

app/views/external_links/_all.html.erb
```diff
@@ -1,5 +1,8 @@
-<%= render partial: 'external_links/psu_digital_collections_links', locals: { psu_digital_collections_links: document.psu_digital_collections_links } if document.psu_digital_collections_links %>
+<% psu_digital_collections_links = document.psu_digital_collections_links
+   access_online_links = document.access_online_links %>
```
nicktiberi, line 2: Do we want to add `hathi_links` to this list of variables, and refactor the partial declaration similarly?

app/models/external_links.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module ExternalLinks
+  def psu_digital_collections_links
```
nicktiberi, line 4: Looking much cleaner! 😎 👍

nicktiberi:
Thanks for cleaning this up, @banukutlu! 🙂

spec/models/access_online_links_spec.rb
banukutlu: oh yes! totally missed this one :) good catch :)

spec/models/access_online_links_spec.rb
banukutlu: yup

spec/models/access_online_links_spec.rb
banukutlu: ok just changed with the latest commit

app/models/external_links.rb
banukutlu: 😎 👍

app/views/external_links/_all.html.erb
banukutlu: yeah good idea 👍

banukutlu: this is done with the latest commit, thanks 👍

PR #717, opened 2021-03-30
672 links refactor 2

banukutlu:
re: #672 

using thumbnail_method and overriding two more partials so we keep the thumbnails linked

todo: need to find a way to make availability wider

### DIFF
--- a/app/views/catalog/_document.html.erb
+++ b/app/views/catalog/_document.html.erb
@@ -1,8 +1,0 @@
-<%# container for a single doc -%>
-<article class="document <%= render_document_class document %> document-position-<%= document_counter %> " data-document-counter="<%= document_counter %>" itemscope itemtype="<%= document.itemtype %>">
-  <div class="row">
-  <%= render_document_partials document,
-                               blacklight_config.view_config(document_index_view_type).partials,
-                               document_counter: document_counter %>
-  </div>
-</article>
--- a/app/views/catalog/_document.html.erb
+++ b/app/views/catalog/_document.html.erb
@@ -1,0 +1,8 @@
+<% # container for a single doc -%>
+<article class="document <%= render_document_class document %> document-position-<%= document_counter%> " data-document-counter="<%= document_counter %>" itemscope itemtype="<%= document.itemtype %>">
+  <div class="row">
+  <%= render_document_partials document,
+    blacklight_config.view_config(document_index_view_type).partials,
+    document_counter: document_counter %>
+  </div>
+</article>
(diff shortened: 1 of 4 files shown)


app/views/catalog/_index_default.html.erb
```diff
@@ -1,32 +1,24 @@
...
-    <%= render partial: 'external_links/all', locals: { document: document } %>
-  </div>
-</div>
+  <%= render partial: 'external_links/all', locals: { document: document } %>
 
-<div class="blacklight-availability">
-  <% unless document.hathi_links && document.hathi_links[:etas_item] && Settings&.hide_etas_holdings -%>
-    <%= render partial: 'index_availability', locals: { document: document } unless Settings&.readonly %>
-  <% end -%>
+  <div class="blacklight-availability">
```
nicktiberi, line 19: @banukutlu - we could move this chunk to the bottom of `app/views/catalog/_document.html.erb` inside the `<article>` tag. That would allow the availability content to span the full width.

app/views/catalog/_document.html.erb
```diff
@@ -0,0 +1,8 @@
+<% # container for a single doc -%>
+<article class="document <%= render_document_class document %> document-position-<%= document_counter%> " data-document-counter="<%= document_counter %>" itemscope itemtype="<%= document.itemtype %>">
+  <div class="row">
+  <%= render_document_partials document,
+    blacklight_config.view_config(document_index_view_type).partials,
+    document_counter: document_counter %>
+  </div>
+</article>
```
nicktiberi, line 8: We could move the availability code here (after the `</div>` and before `</article>`:

```
  <div class="blacklight-availability">
    <% unless document.hathi_links && document.hathi_links[:etas_item] && Settings&.hide_etas_holdings -%>
        <%= render partial: 'index_availability', locals: { document: document } unless Settings&.readonly %>
    <% end -%>
  </div>
  ```

nicktiberi (commented):
Let me know what you think! There may be a cleaner way to do it, but this would allow us to maintain the link in the thumbnail while allowing the availability stuff to stay full-width.

app/views/catalog/_index_default.html.erb
banukutlu: @nicktiberi would you mind giving this a try? It could work 👍

app/views/catalog/_document.html.erb
banukutlu: @nicktiberi would you mind giving this a try as well?

nicktiberi:
@banukutlu and @awead - this is ready for re-review. I cleaned up the rendering a bit so that the margins and everything look the same as in #705, but the thumbnail link still works in the index view.

banukutlu:
works nicely! thanks for wrapping the rest up @nicktiberi

PR #718, opened 2021-04-05
Availability Testing Improvements

nicktiberi:
Re: #719 

I've incorporated the VCR gem in the availability tests to mock the SIRSI responses. This will allow us to have reliable/repeatable tests, and we will no longer actually have to make requests to the production instance of Symphony in our test suite.

**NOTE: ** we will be adding some new records to the sample dataset:
- 3500414
- 33183518
- 6962697


nicktiberi:
@cdmo and @banukutlu  - let me know what you think! Could probably be cleaned up some more, but I think it might have potential.

app/controllers/availability/availability_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'net/http'
+
+module Availability
+  class AvailabilityController < ActionController::API
+    def sirsi_data
+      sirsi_url = 'https://cat.libraries.psu.edu:28443/symwsbc/rest/standard/lookupTitleInfo' +
```
cdmo, line 8: I get that this is a proof of concept, but, one quick thing I think would be to move this url into an application config setting in https://github.com/psu-libraries/psulib_blacklight/blob/main/config/settings.yml

Another thing that could be done is something similar to how we did things in myaccount, with a Service class to act as a client for the web service. Could do that too, like place the web service interaction logic there and have this controller just call it. Certainly other ways to organize the code too, like query objects or concerns... In Rails dev worlds the phrase "Fat Controller" comes up a lot as a thing to avoid but there isn't consensus always on how to make it "skinnier"

Just some thoughts. Overall I think this refactor is well worth it and will make the availability code much simpler

nicktiberi: Thanks for the feedback! I moved the URL out into the settings file. I'll look into the service class thing as well.

I think now I will just stub/mock all the requests with `stub_sirsi_request` to get the tests passing and then we can go from there.

banukutlu (dismissed):
yeass! lgtm 🎉  once we have the sample file updated, it looks good to go to me 👏

nicktiberi:
@banukutlu 🟢 🟢 🟢  😎

banukutlu:
> @banukutlu 🟢 🟢 🟢 😎

yeass 🚀 👏

app/javascript/availability/index.js
```diff
@@ -554,12 +551,14 @@ const availability = {
             const day = dueDate.toLocaleDateString('en-us', {
                 year: 'numeric',
                 month: 'numeric',
-                day: 'numeric'
+                day: 'numeric',
+                timeZone: 'America/New_York'
```
banukutlu, line 555: 👍

banukutlu (approved):
🚢 :shipit: 🚀  

thx so much for all this awesome work @nicktiberi!

PR #724, opened 2021-04-13
Disable search history

### LINKED ISSUE #260: Drop searches table
Saved searches aren't needed anymore, we need to remove the table and disable this feature entirely. New migration.

If we decide we still need it then we need a scheduled task to clear the table periodically.

awead:
Blacklight's search session depends on loading recent searches from the Searches table. Completely dropping the table would involve a lot of overriding in `Blacklight::SearchContext`. Additionally, we'd be inheriting some future technical debt whenever we upgraded. 

Instead of getting rid of the table, we disable the search history feature in Blacklight by making the Search model readonly. This allows all of Blacklight to keep on working like nothing's changed, except records are never stored to the database.

A user's session uses a readonly Search record, allowing all of the current behaviors to function, with minimal overrides to Blacklight. An additional override to the Search model ensures that any additional or future attempts to save records will not occur.

Fixes #260

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ group :development do
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+  gem 'web-console', '>= 3.3.0'
 end
 
 group :development, :test do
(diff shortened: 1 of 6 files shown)


Gemfile
```diff
@@ -32,6 +32,7 @@ group :development do
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+  gem 'web-console', '>= 3.3.0'
```
banukutlu, line 35: 😎

banukutlu (approved):
kewl! works nicely 👌  thanks @awead!

PR #729, opened 2021-04-20
Revert disable search history

awead:
Reverts the commit that removed the search history dependency, which _broke_ searching! This will be fixed again when we switch to Redis.


Gemfile
```diff
@@ -32,7 +32,6 @@ group :development do
   gem 'listen', '>= 3.0.5', '< 3.2'
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
-  gem 'web-console', '>= 3.3.0'
```
banukutlu, line 35: @awead this could have stayed though ;) guessing you will add it with the other PR.

banukutlu (approved):
👍

PR #730, opened 2021-04-20
Use Redis to store searches

### LINKED ISSUE #260: Drop searches table

awead:
Use Redis to store search sessions instead of the `searches` table.

## Changes

* configures Redis for use with the application
* overrides `Blacklight::SearchContext` to store a user's previous searches in Redis for the duration of their session
* removes `searches` table
* renders Blacklight's search history features unusable
    * users cannot permanently save their searches
    * we don't allow this in the UI anyway    

## Details

When performing searches in the catalog, the following Redis keys will be saved:

```
> keys *
1) "search:af38b0d75d982f528d50661fcc07d354"
2) "search:42953f6bde8044dbfe47c2b53c5411ee"
3) "search:1a2f76e87dc95dc27c76cea156281122"
4) "searchhistory:801cd474dcf91cc35a3d43fe18a58ce7"
> lrange searchhistory:801cd474dcf91cc35a3d43fe18a58ce7 0 -1
1) "search:af38b0d75d982f528d50661fcc07d354"
2) "search:42953f6bde8044dbfe47c2b53c5411ee"
3) "search:1a2f76e87dc95dc27c76cea156281122"
> get search:af38b0d75d982f528d50661fcc07d354
"{\"f\":{\"format\":[\"Article\"]},\"controller\":\"catalog\",\"action\":\"index\"}"
```

For each user sessions, there is a `searchhistory:*` key with an array of `search:*` keys each representing one of the user's searches. The search key will hold a json string of the search parameters for a given search. Keys are set to expire after 24 hours, but this can configured via the `REDIS_TTL` environment variable.

Fixes #260


lib/psulib_blacklight/redis_config.rb
```diff
@@ -0,0 +1,40 @@
...
+    def host
+      ENV.fetch('REDIS_HOST', 'localhost')
+    end
+
+    def port
+      ENV.fetch('REDIS_PORT', 6379)
+    end
+
+    def database
+      ENV.fetch('REDIS_DATABASE', 0)
```
whereismyjetpack, line 14: I think we should pull values out of `Settings` here since we are using the config gem. this will allow configuration to be put in the environment, or via config gem, this will allow deployment on linux and containers. maybe something like Settings.redis.host || ENV.fetch?

whereismyjetpack (changes requested):

whereismyjetpack (approved):
There was a typo in Redis.new, uri != url (we use uri in config other places, so i didn't rename that) Deployed to QA for testing

whereismyjetpack:
Production server should be ready, dependant on https://github.com/psu-libraries/psulib_blacklight_deploy/pull/110

awead:
@whereismyjetpack 
> There was a typo in Redis.new, uri != url (we use uri in config other places, so i didn't rename that) Deployed to QA for testing

This seems to have broken the build. Did you test this locally? It's failing for me, but the change seems correct.

awead:
@whereismyjetpack ok, figured it out. The your correction is correct, the tests just need to be updated. I don't know if it deployed to QA since the tests failed.

whereismyjetpack:
@awead  good find, I'll re-deploy and we should be good. I thought the failure was CC related, whoopsies!

awead:
@whereismyjetpack yeah, me too. It's hard to tell with the output. I miss the red colored text for failures!

banukutlu:
So. to be able to use redis locally, I used

```config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_URL") { "redis://localhost:6379/3" } }``` to here https://github.com/psu-libraries/psulib_blacklight/blob/redis-searches/config/environments/development.rb#L23.

@awead how do you run redis locally?

File: spec/models/search_spec.rb [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+
+  describe '#save' do
+    let(:params) { { q: 'foo', fl: ['bar'] } }
+    let(:search) { described_class.new(query_params: params) }
+
+    before { search.save }
+
+    it 'saves the key/value pair to Redis as a json string with an expiration time' do
+      expect(redis_connection.get(search.id)).to eq(params.to_json)
+      expect(redis_connection.ttl(search.id)).to be >= 86_400
```
banukutlu, line 103:
```suggestion
      expect(redis_connection.ttl(search.id)).to be >= Settings.redis.sessions.ttl
```
guess we can use this value from config as well?

banukutlu (commented):
looks great! just had a question about using redis locally, depending on that maybe an update to readme.

awead:
@banukutlu I added a bit to the readme. All you need to do is have redis running locally, or update settings.yml.

banukutlu:
> @banukutlu I added a bit to the readme. All you need to do is have redis running locally, or update settings.yml.

thanks Adam 👍

banukutlu (dismissed):

banukutlu (approved):

Issue #720, opened 2021-04-05
Surface "Report an Issue" button / add form


ruthtillman:
On individual item page next to "Bookmark" and "Share" buttons add "Report an Issue" button (on the right). This button would then send users to a web form which would allow them to submit a report to the team in cataloging that handles fixing records. 

Form fields, label is bolded:

- Autopopulate: **Title** (solr: `title_245ab_tsim`)
- Autopopulate: **Record number** (solr: `id`)
- **Comment** (larger text box so that a person can write a longer comment)
- **Email address** add: "(optional, if you'd like a reply)"

buttons: Send / Cancel

email to: `catalog-record-issue@PennStateOffice365.onmicrosoft.com`

ruthtillman:
Note: the Oops form has Call #, Copy #, and item ID -- but these are all populated from the first available copy... which makes them not useful for specifics. If it has 4 copies, it'll only do 1 copy. So we're not going to use them for now. If there's need, we'll have to sort out what exactly we need and why and can enhance the form.

ruthtillman:
Add URL to catalog with prefix: `https://catalog.libraries.psu.edu/catalog/$id`

banukutlu:
tested on QA, flash msg displaying on the dialog.

<img width="860" alt="Screen Shot 2021-05-04 at 9 23 10 PM" src="https://user-images.githubusercontent.com/6477867/117088376-4b5ee100-ad20-11eb-8fc1-5bec0f3b7531.png">

PR #732, opened 2021-04-30
Add a "Report an Issue" form to catalog records

nicktiberi:
Re: #720.

This PR adds a "Report an Issue" button to the record toolbar. When clicked, a modal appears, showing a form that a user can fill out to report an issue. An email is then sent to `report_issue_email_recipient` (which in production + QA, should be `catalog-record-issue@PennStateOffice365.onmicrosoft.com`).

### New Button
Screen Shot 2021-04-30 at 11 28 36 AM

### Report an Issue Form/Modal
Screen Shot 2021-04-30 at 11 29 04 AM

### Form Submission Success
Screen Shot 2021-04-30 at 11 29 14 AM

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -5,6 +5,7 @@ class CatalogController < ApplicationController
   include BlacklightRangeLimit::ControllerOverride
   include Blacklight::Catalog
   include Blacklight::Marc::Catalog
+  include ::ReportIssue
 
   def index
     cache_key = nil
@@ -64,11 +65,13 @@ class CatalogController < ApplicationController
   configure_blacklight do |config|
     # Controls the document actions (also called "tools"), note that blacklight_marc adds refworks and endnote
     config.add_show_tools_partial(:email, callback: :email_action, validator: :validate_email_params, html_class: 'dropdown-item')
+    config.add_show_tools_partial(:report_issue, callback: :report_issue_action, validator: :validate_report_issue_params)
     # TODO: hide SMS action for now, should be enabled when fixed
     # config.add_show_tools_partial(:sms, if: :render_sms_action?, callback: :sms_action, validator: :validate_sms_params, html_class: 'dropdown-item')
     config.show.document_actions.refworks.html_class = 'dropdown-item'
     config.show.document_actions.endnote.html_class = 'dropdown-item'
     config.show.document_actions.delete_field('librarian_view') # removing something added by blacklight_marc
+    config.show.document_actions.delete_field('report_issue') # hide the 'Report an Issue' action from the Share dropdown
 
     # default advanced config values
     config.advanced_search ||= Blacklight::OpenStructWithHashAccess.new
--- a/app/models/record_mailer.rb
+++ b/app/models/record_mailer.rb
@@ -17,6 +17,15 @@ class RecordMailer < ApplicationMailer
     mail(to: details[:to], subject: subject)
   end
 
+  def email_report_issue(title:, record_number:, comment:, email: '')
+    @title = title
+    @record_number = record_number
+    @comment = comment
+    @email = email
+
+    mail(to: Settings.report_issue_email_recipient, subject: t('blackcat.report_issue.email.subject'))
+  end
+
   def sms_record(documents, details, url_gen_params)
     @documents      = documents
     @url_gen_params = url_gen_params
--- a/app/views/catalog/_report_issue_form.html.erb
+++ b/app/views/catalog/_report_issue_form.html.erb
@@ -1,0 +1,59 @@
+<%= form_tag url_for(controller: controller_name, action: "report_issue"),
+             data: { blacklight_modal: 'trigger' },
+             id: 'report_issue_form',
+             class: "modal_form",
+             method: :post do %>
+  <div class="modal-body">
+    <%= render '/shared/flash_msg' %>
+
+    <div class="form-group row">
+      <label class="control-label col-sm-3" for="to">
+        <%= t('blackcat.report_issue.form.record_title') %>
+      </label>
+      <div class="col-sm-9">
+        <%= text_field_tag :title, @documents.first['title_display_ssm'], class: 'form-control', :readonly => true %>
+      </div>
+    </div>
+
+    <div class="form-group row">
+      <label class="control-label col-sm-3" for="to">
+        <%= t('blackcat.report_issue.form.record_number') %>
+      </label>
+      <div class="col-sm-9">
+        <%= text_field_tag :record_number, @documents.first.id, class: 'form-control', :readonly => true %>
+      </div>
+    </div>
+
+    <div class="form-group row">
+      <label class="control-label col-sm-3" for="message">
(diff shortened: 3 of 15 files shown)


config/settings.yml
```diff
@@ -1,5 +1,6 @@
 hold_button_url: http://cat.libraries.psu.edu/cgi-bin/catredirpg?C=
 matomo_id: 10
+report_issue_email_recipient: test@test.com
```
nicktiberi, line 3: @banukutlu - I talked to @awead about this and he mentioned preferring to store info like this in environment variables which could be pulled down from Vault. Is the catalog storing any data in environment variables right now, or are all the settings just in `settings.yml`?

Not sure if this is something that needs to be changed before this PR gets merged, but wanted to bring it up.

app/views/catalog/_report_issue_form.html.erb
```diff
@@ -0,0 +1,59 @@
+<%= form_tag url_for(controller: controller_name, action: "report_issue"),
```
nicktiberi, line 1: @awead - the linter is crapping out on this file with the error:

```
Exception occured when processing: app/views/catalog/_report_issue_form.html.erb
If this file cannot be processed by erb-lint, you can exclude it in your configuration file.
begin_pos location 2052 is out of range for document of size 2016
```

I'm not really sure why `begin_pos` is 2052 instead of 0. Have you seen something like this before?

FWIW, I replaced the contents of this file with the contents of`_email_form.html.erb` from the Blacklight gem, and it failed with the same error, but different numbers, due to the different file size.

config/settings.yml
awead: We use Vault in Scholarsphere, which just sets environment variables. I'm _assuming_ there's something similar going on here. The gist was, don't put a real email in the repo here, and the deployment (QA, prod, etc.) would set this to the real email where you want things to go.

app/models/record_mailer.rb
```diff
@@ -17,6 +17,16 @@ def email_record(documents, details, url_gen_params)
     mail(to: details[:to], subject: subject)
   end
 
+  def email_report_issue(_documents, details, url_gen_params)
```
awead, line 20: `_documents` isn't being used in this method. Rubocop, I bet, put that underscore in there automatically. That's a Ruby-ism for indicating that a parameter is being passed in but not used. So you should be able to delete it and just have:

``` ruby
def email_report_issue(details, url_gen_params)
...
```

app/controllers/catalog_controller.rb
```diff
@@ -61,14 +61,40 @@ def show
     super
   end
 
+  def report_issue_action(documents)
+    mail = RecordMailer.email_report_issue(documents, {
```
awead, line 65: We're actually not using the `documents` variable here at all. When we do the `config.add_tools_partial` thing, Blacklight is generating a method called `report_issue`. That method that sets the `@documents` variable that you're using in your form. This is 👍 . But, when we make the callback to `report_issue_action`, because Blacklight has _generated_ the `report_issue` method, it's passing `@documents` into our callback. We can't change that...unless we write our own `report_issue` method, but nah... we'll just discard the documents parameter here, and only pass what the email needs:

``` ruby
def report_issue_action(_documents)                                                                                      
    mail = RecordMailer.email_report_issue(                                                                                
      title: params[:title],                                                                                               
      record_number: params[:record_number],                                                                               
      comment: params[:comment],                                                                                           
      email: params[:email],                                                                                               
      url_options: url_options)
```

This will change the method signature of `email_report_issue` slight because we're using keyword arguments:

``` ruby
def email_report_issue(title:, record_number:, comment:, email:, url_gen_params:)
```

app/views/catalog/_report_issue_form.html.erb
awead: Weird. Yeah, I've seen this happen sometimes. I just add the file to the exclude block in `.erblint.yml`:

```
exclude:
- '**/app/views/catalog/_report_issue_form.html.erb'
```

spec/models/record_mailer_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+require 'rails_helper'
+
+RSpec.describe RecordMailer do
+  before do
+    allow(described_class).to receive(:default).and_return(from: 'no-reply@libraries.psu.edu')
+    SolrDocument.use_extension(Blacklight::Document::Email)
+    SolrDocument.use_extension(Blacklight::Document::Sms)
+    document = SolrDocument.new
+    @documents = [document]
```
awead, line 11: you shouldn't need this now.

awead (changes requested):
Looks good! I hope this wasn't too annoying. In reviewing it, I think I realized that we probably would have been better off just not doing it this way to begin with, but whatever, it's working now! If you want to give it another go and do it the non-Blacklight, Rails standard way, I'm game. Otherwise, there's a few changes to get rid of `documents` and then it should be good to go.

app/controllers/catalog_controller.rb
```diff
@@ -61,14 +61,38 @@ def show
     super
   end
 
+  def report_issue_action(_documents)
```
nicktiberi, line 64: @awead - I don't love having this just chilling in `catalog_controller`. It seems like it's just cluttering the catalog controller and should maybe be in a separate file.

The main blacklight project puts `email_action` and `validate_email_params` in `app/controllers/concerns/blacklight/catalog.rb`. 

Should we do something similar? Or do you think it's fine to just leave this here? I defer to your Rails expertise. 🙂

nicktiberi:
@awead - thanks for reviewing! If you're ok with it, I think for now maybe we can just go the Blacklight customization route. Now that I have it implemented, it does _kinda_ make sense to me. 😁 

If we need to do any more custom actions in the future, maybe we can do the standard Rails approach for the new action, and then revisit this?

app/controllers/catalog_controller.rb
```diff
@@ -61,14 +61,38 @@ def show
...
+  end
+
+  def validate_report_issue_params
+    if params[:comment].blank?
+      flash[:error] = t('blackcat.report_issue.errors.comment')
+    end
+
+    flash[:error].blank?
+  end
+
```
nicktiberi, line 85: @awead - I don't love having this just chilling in `catalog_controller`. It seems like it's just cluttering the catalog controller and should maybe be in a separate file.

The main blacklight project puts `email_action` and `validate_email_params` in `app/controllers/concerns/blacklight/catalog.rb`. 

Should we do something similar? Or do you think it's fine to just leave this here? I defer to your Rails expertise. 🙂

awead: Yep. That’s a great idea. You could create a module under concerns and include it in the controller.

nicktiberi:
@awead - should be ready for re-review. I moved those two methods into a concern and also added the record URL to the email as Ruth suggested.

awead (approved):

config/settings.yml
banukutlu: We are using config gem to handle these, production.local.yml or settings.local.yml where we can add the real email.

Issue #736, opened 2021-05-05
Add prefix, label and notes for external links


banukutlu:
https://github.com/psu-libraries/psulib_traject/issues/277

https://github.com/psu-libraries/psulib_blacklight/wiki/URL-Fields#link-label-and-note-fields

`3: <a href="u">y</a>, z`

ruthtillman:
🤔 so it looks very good except for one issue from the data in notes. Sometimes the subfield z ends with a ":" - example catkey `32173572` ...  can we trim any final colon off the "notes" field in

```json
"full_links_struct":["{"prefix":"","text":"doi.org","url":"https://doi.org/10.3998/mpub.17577","notes":"Open Access:"}"]
```

Example of publication which contains colons but is fine. `33848484`

I think this is because sometimes the notes are put in the record before the URL, but because they can be very long, that's not something we want to do.

PR #739, opened 2021-05-11
adding prefix and notes to external links

banukutlu:
#736


nicktiberi (approved):
Tested a few records after reindexing locally, and things look good to me! 🥇

PR #742, opened 2021-05-13
Improve checks for PSU Digital Collections links URLs

nicktiberi:
Re: #672.

Tightens up our checks for which content is actually part of PSU Digital Collections.

Content is part of PSU Digital Collections if it contains one of the following URLs:

- digital.libraries.psu.edu
- libraries.psu.edu/collections
- collection[x].libraries.psu.edu


banukutlu (approved):
👌

PR #743, opened 2021-05-13
External Links: trim trailing colon from 'notes' field

nicktiberi:
Re: #736.

Any trailing colons on the notes field are now trimmed if present.

### Before:
Screen Shot 2021-05-13 at 3 09 45 PM

### After:
Screen Shot 2021-05-13 at 3 09 52 PM


banukutlu (approved):
👌

PR #746, opened 2021-05-19
need to explicitly define facet queries 

banukutlu:
see here `https://github.com/projectblacklight/blacklight/issues/2401`


nicktiberi (dismissed):
Great find! 👌

nicktiberi (approved):
🎉

PR #748, opened 2021-05-25
using components for constraints instead of overriding partials

banukutlu:
https://github.com/projectblacklight/blacklight/wiki/Configuration---Results-View#using-document_component


awead (dismissed):
I wish there was a better way to do this, but you've done it the best way possible given the setup. Unfortunately, It seems terribly convoluted, yet I can't see a cleaner way. I think the best alternative would be to completely replace Blacklight's constraint system with our own component-base one, that removes the constraints calling other constraints, and other things, but that's out of scope for now. I suspect later releases will clean this up a bit and this is a first step towards getting away from the helper-based system.

banukutlu:
@awead this is ready for review again after adding some specs. thanks!

spec/components/psul_constraint_layout_component_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  describe 'for simple display' do
+    let(:params) do
+      { label: 'my label', value: 'my value' }
+    end
+
+    it 'renders label and value' do
+      expect(rendered).to have_selector('span.applied-filter.constraint') do |s|
+        expect(s).to have_css('span.constraint-value')
```
awead, line 17: I don't think this block is doing what you think it's doing. I believe what is happening is that after line 16, the expect statement evaluates and the block is ignored. If you put in `expect(s).to have_css('span.garbage-value')` it will still pass.

The easiest thing to do would just be to put them all inline:

```ruby
      expect(rendered).to have_selector('span.applied-filter.constraint')                                                  
      expect(rendered).to have_css('span.constraint-value')                                                                
      expect(rendered).not_to have_css('a.constraint-value')                                                               
      expect(rendered).not_to have_css('span.btn.btn-outline-secondary.constraint-value')                                  
      expect(rendered).to have_selector 'span.filter-name', text: 'my label'                                               
      expect(rendered).to have_selector 'span.filter-value', text: 'my value'
```

I was looking around for something like `within('span.applied-filter.constraint') do` which is what we do in feature tests, but rendered would need to be a `Capybara::Node::Simple` or something. Probably not worth getting into now, but something we could refactor later.

awead (changes requested):
@banukutlu Looks good. You'll need to change all the ` do |s|` blocks. See my first comment.

spec/components/psul_constraint_layout_component_spec.rb
banukutlu: ah yeah thanks for catching this, I had it as Capybara::Node::Simple then removed it, I will put that back and check

banukutlu: hah I thought 
```
subject(:rendered) do
    Capybara::Node::Simple.new(render_inline(described_class.new(**params)))
end
```
would work but same behavior 🤔  I mean it passes for any value.

guess I will just put them inline as you suggested, unless you have suggestion for how to get `Capybara::Node::Simple` to work.

awead: I'd say leave it for now and we can revisit later.

banukutlu: yup, just converted them to inline 👍

awead (approved):

Issue #696, opened 2021-02-15
Rename Open Access Label / add tooltip


ruthtillman:
Per Scholarly Communications Outreach Librarian and meeting with selected of public-facing/outreach librarians/staff, we are going to be renaming the Open Access label "Free to Read" and adding a tooltip. 

Tooltip text: `These materials don't require a Penn State login`

Add tooltip to Online facet as well: `These materials may require a Penn State login`

PR #753, opened 2021-06-08
Add tooltips to the "Online" and "Free to Read" facet items

nicktiberi:
Re: #696.

Tooltips (via the HTML `title` attribute) will now appear for "Online" and "Free to Read" facet items. This PR makes it easy to add new tooltips to other facet items by adding new elements to the `FACET_TOOLTIPS` hash.

Screen Shot 2021-06-08 at 10 34 39 AM

### DIFF
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -55,4 +55,9 @@ module FacetsHelper
     pivot.each { |field| in_params = true if params[:f] && params[:f][field] }
     in_params
   end
+
+  def facet_item_component_class(facet_config)
+    default_component = facet_config.pivot ? Blacklight::FacetItemPivotComponent : PsulFacetItemComponent
+    facet_config.fetch(:item_component, default_component)
+  end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -39,6 +39,10 @@ en:
         restricted_access_text: 'Search inside at HathiTrust'
         etas_text: 'Check out digital copy through HathiTrust'
         etas_additional_text: 'By special arrangement in response to the COVID-19 pandemic, patrons may access this resource online through the HathiTrust Emergency Temporary Access Service.'
+      facet_tooltips:
+        access_facet:
+          free_to_read: 'These materials don''t require a Penn State login'
+          online: 'These materials may require a Penn State login'
       report_issue:
         form:
           title: 'Report an Issue'
--- a/app/components/psul_facet_item_component.rb
+++ b/app/components/psul_facet_item_component.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+class PsulFacetItemComponent < Blacklight::FacetItemComponent
+  def initialize(psul_facet_item:, wrapping_element: 'li', suppress_link: false)
+    super(facet_item: psul_facet_item, wrapping_element: wrapping_element, suppress_link: suppress_link)
+  end
+
+  ##
+  # Standard display of a facet value in a list. Used in both _facets sidebar
+  # partial and catalog/facet expanded list. Will output facet value name as
+  # a link to add that to your restrictions, with count in parens.
+  #
+  # @return [String]
+  # @private
+  def render_facet_value
+    tag.span(class: 'facet-label') do
+      link_to_unless(@suppress_link, @label, @href, class: 'facet-select', title: facet_tooltip(@label))
+    end + render_facet_count
+  end
+
+  ##
+  # Standard display of a SELECTED facet value (e.g. without a link and with a remove button)
+  # @see #render_facet_value
+  #
+  # @private
+  def render_selected_facet_value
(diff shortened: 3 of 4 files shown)


app/components/blacklight/facet_item_component.rb
```diff
@@ -0,0 +1,129 @@
...
+    end
+
+    ##
+    # Standard display of a SELECTED facet value (e.g. without a link and with a remove button)
+    # @see #render_facet_value
+    #
+    # @private
+    def render_selected_facet_value
+      tag.span(class: 'facet-label') do
+        tag.span(@label, class: 'selected', title: facet_tooltip(@label)) +
```
nicktiberi, line 86: Added `title: facet_tooltip(@label)`

app/components/blacklight/facet_item_component.rb
```diff
@@ -0,0 +1,129 @@
...
+      def facet_tooltip(key)
+        FACET_TOOLTIPS.dig(@facet_item.facet_field, key)
+      end
+
+      FACET_TOOLTIPS = {
+        'access_facet' => {
+          'Free to Read' => I18n.t('blackcat.facet_tooltips.access_facet.free_to_read'),
+          'Online' => I18n.t('blackcat.facet_tooltips.access_facet.online')
+        }
+      }.freeze
```
nicktiberi, line 127: Added all of this.

app/components/blacklight/facet_item_component.rb
```diff
@@ -0,0 +1,129 @@
...
+    ##
+    # Standard display of a facet value in a list. Used in both _facets sidebar
+    # partial and catalog/facet expanded list. Will output facet value name as
+    # a link to add that to your restrictions, with count in parens.
+    #
+    # @return [String]
+    # @private
+    def render_facet_value
+      tag.span(class: 'facet-label') do
+        link_to_unless(@suppress_link, @label, @href, class: 'facet-select', title: facet_tooltip(@label))
```
nicktiberi, line 75: Added `title: facet_tooltip(@label)`

app/components/blacklight/facet_item_component.rb
```diff
@@ -0,0 +1,129 @@
+# frozen_string_literal: true
```
nicktiberi, line 1: This file is almost entirely copied from the existing `FacetItemComponent` in Blacklight, with a few additions noted in comments below. 

If there is a better way to do this than overriding the component file (subclassing the component, maybe?), please let me know. It seems not great to me to have to override everything in this component just to add some small functionality, so maybe I'm missing something here! 🤓

awead (changes requested):
If possible, I'd really like to find a way we can make this overrides without copying and entire class of code for just a few lines' change. I realize this may not be possible, but it'd really be worth finding a way. We inherit a lot of technical debt this way, and I'm also seeing drops in test coverage.

app/helpers/facets_helper.rb
```diff
@@ -55,4 +55,9 @@ def pivot_facet_field_in_params?(pivot)
     pivot.each { |field| in_params = true if params[:f] && params[:f][field] }
     in_params
   end
+
+  def facet_item_component_class(facet_config)
+    default_component = facet_config.pivot ? Blacklight::FacetItemPivotComponent : PsulFacetItemComponent
```
nicktiberi, line 60: @awead - By overriding this method, we can user our custom component and subclass the default component, avoiding all that code duplication of my previous approach.

config/locales/en.yml
```diff
@@ -39,6 +39,10 @@ en:
         restricted_access_text: 'Search inside at HathiTrust'
         etas_text: 'Check out digital copy through HathiTrust'
         etas_additional_text: 'By special arrangement in response to the COVID-19 pandemic, patrons may access this resource online through the HathiTrust Emergency Temporary Access Service.'
+      facet_tooltips:
+        access_facet:
+          free_to_read: 'These materials don''t require a Penn State login'
```
banukutlu, line 44:
```suggestion
          free_to_read: 'These materials don't require a Penn State login'
```

nicktiberi: @banukutlu - the double apostrophe (`''`) was intentional - it is an escape for the single apostrophe because this text is in single quotes. Alternatives would be to use double quotes or no quotes:

```
free_to_read: These materials don't require a Penn State login
free_to_read: "These materials don't require a Penn State login"
```

Do you have a preference?

banukutlu: Ahh my bad sorry!  What you had is great, I rather single apostrophe here as well, no need change 👌

awead (approved):

config/locales/en.yml
awead: For future reference, the quotes aren't always necessary in yaml, but if it needs them, you can use double-quotes. If you happen to need double-quotes in an already double-quoted string, then `"He said \"What!?\""`

Issue #740, opened 2021-05-12
Experiment with Public Note Display


ruthtillman:
The WebServices response includes a publicNote in some responses which can have special information about items, especially for Special Collections. Let's try displaying it, but then I'll have to do a bunch of testing to be sure we don't have something inappropriate in these fields in non-SCL records.

Sample catkeys:

- `6962697`
- `20499047`

This is not a rush, though if @nicktiberi has time to experiment with it, that would be great.

Note: the publicNote is at the Item level and would be displayed just below that individual item in Availability, similarly to other data that might be displayed about that item.

PR #757, opened 2021-06-10
Availability: Add public note tooltip

nicktiberi:
Re: #740.

For items which have a `publicNote`, we now display an info icon with a tooltip.

image

### DIFF
--- a/spec/features/availability_spec.rb
+++ b/spec/features/availability_spec.rb
@@ -66,6 +66,18 @@ RSpec.feature 'Availability', type: :feature do
       expect(page).to have_selector 'strong', text: /Being acquired by the library/
     end
 
+    xit 'that has a public note' do
+      visit '/?utf8=✓&search_field=all_fields&q=6962697'
+      click_button('View Availability')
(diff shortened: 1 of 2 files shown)


spec/features/availability_spec.rb
```diff
@@ -66,6 +66,18 @@
       expect(page).to have_selector 'strong', text: /Being acquired by the library/
     end
 
+    xit 'that has a public note' do
```
nicktiberi, line 69: @banukutlu - I'm going to revisit the work I started a while ago in improving these unit tests. We're starting to add even more pending specs. 😭

banukutlu: ahhhh yeah! availability specs always 😬

banukutlu (approved):
lgtm 👍  

just to note availability js is getting even bigger, maybe we should consider breaking it down at some point.

nicktiberi:
> lgtm 👍
> 
> just to note availability js is getting even bigger, maybe we should consider breaking it down at some point.

@banukutlu - completely agree. Maybe after we get the testing working better we can chat about refactoring the availability code.

PR #768, opened 2021-06-22
kubernetes and docker-compose support

whereismyjetpack:
- migrates to circleCI so we can use our CI/CD process for deployments
- adds docker-compose support for local development
- creates an `initalize_config` rake task, which is an idempotent version of `create_collection`
- adds username and password support to solr_manager (this is needed for newer versions of solr to use the config API)
- breaks out solr config into protocol, host, port, collection parts with sane defaults. (to more easily configure via environment
- adds ENV variable support via ERB, and via `config`
- updates README with docker specific tasks. 

TODO
- wire in traject better. 
My current thought for this is to have a `traject` service in the `docker-compose` file that we can use to seed the solr index? 
- bring solr_manager in line across traject and psulib_blacklight.
- clean up legacy rake tasks that shell out to `docker`


whereismyjetpack:
TODO isn't a "task list" per-say, just thoughts I had while working on this, that I'd be fine with scrapping, or evolving.

config/database.yml
```diff
@@ -25,8 +25,9 @@ test:
 production:
   adapter: mysql2
   encoding: utf8
+  host: <%= ENV.fetch("MYSQL_HOST") { "127.0.0.1" } %>
```
awead, line 28: @whereismyjetpack Use `{ }` with fetch only if you need to execute the block. If you're just wanting a default value, then `ENV.fetch("MYSQL_HOST", "127.0.0.1")`  See https://ruby-doc.org/core-2.7.0/Hash.html#method-i-fetch

config/blacklight.yml
```diff
@@ -1,12 +1,28 @@
 development:
   adapter: solr
-  url: <%= "#{Settings.solr.url}/solr/#{Settings.solr.collection}" %>
+  <% if Settings.solr.password %>
```
awead, line 3: What about just `SolrConfig.new.url` here and it figures it out for us. We can also copy what we did in Scholarsphere, if that works: https://github.com/psu-libraries/scholarsphere/blob/develop/config/application.rb#L48

whereismyjetpack: `SolrConfig` isn't available here, not sure why.

awead: Try putting `require 'psulib_blacklight/solr_config'` in `application.rb` but inside the Application class definition, like here https://github.com/psu-libraries/scholarsphere/blob/62d9613d503b29ffef05e113b8a032bc6943c793/config/application.rb#L27

whereismyjetpack: I've added a `query_url` method for this use case, and simplified the blacklight config. i hope it suffices

docker-compose.yml
```diff
@@ -0,0 +1,50 @@
...
+      context: . 
+      target: dev
+    volumes:
+    - bundle-data:/app/vendor/bundle
+    - node-data:/app/node_modules
+    - type: bind
+      source: ./
+      target: /app/
+    ports:
+      - "3000:3000"
```
banukutlu, line 34: @whereismyjetpack  
```
ports:
      - "${APP_PORT:-3000}:3000"
```
can we do the same we did for myaccount here?

lib/psulib_blacklight/solr_manager.rb
```diff
@@ -55,6 +55,15 @@ def last_incremented_collection
       collections_with_prefix.max_by { |collection| version_number collection }
     end
 
+    def initialize_collection
```
banukutlu, line 58: @whereismyjetpack 

would you mind adding a spec for this new method, we will get the coverage higher ;) 

https://github.com/psu-libraries/psulib_blacklight/blob/main/spec/lib/psulib_blacklight/solr_manager_spec.rb

whereismyjetpack: I can work on this tomorrow.

docker-compose.yml
whereismyjetpack: This should be fixed up

lib/psulib_blacklight/solr_manager.rb
whereismyjetpack: I'm afraid i might need help on this. I can not, for the life of me, ever wrap my head around rspec. 

Given the following code, the create_collection method does get exersized, i see it's `puts` output in the console, but my setup is completely broken. I'm hoping someone can look at this and point at the exact thing i'm doing wrong. 


```
  describe '#initialize_collection' do
    context 'when collection does not exist' do 
      let(:spy_solr_manager) { instance_spy(described_class) }

      before do 
        stub_request(:get, "#{config_obj.url}/solr/admin/collections?action=LIST")
          .to_return(status: 200, body: '{"responseHeader":{"status":0, "QTime":11}, "collections":[""]}')
      end

      it 'does does not add a new collection' do 
        solr_manager.initialize_collection
        expect(spy_solr_manager.initialize_collection).to have_received(:create_collection)
      end
    end
  end
  ```

  Sorry!

Comment (awead) [reply lib/psulib_blacklight/solr_manager.rb:58]: are you still blocked on this?

File: README.md [hunk 5]
```diff
@@ -54,6 +54,41 @@ Note: we have only developed on Macs thus far.
...
+
+Start up the application with all it's dependencies 
+```
+docker-compose up --build -d
+```
+
+Initialize database, solr
+```
+docker-compose exec web bundle exec db:migrate
+docker-compose exec web bundle exec solr:initialize_collection
```
banukutlu, line 67:
```suggestion
docker-compose exec web bundle exec rake db:migrate
docker-compose exec web bundle exec rake solr:initialize_collection
```

banukutlu, line 67:
```suggestion
docker-compose exec web bundle exec rake db:migrate
docker-compose exec web bundle exec rake solr:initialize_collection
```

lib/psulib_blacklight/solr_manager.rb
whereismyjetpack: @banukutlu  is looking at it for me

whereismyjetpack: We worked on this and concluded that spies probably aren't needed @awead thoughts?

awead: LGTM! 👍

Dockerfile
```diff
@@ -0,0 +1,83 @@
+FROM harbor.k8s.libraries.psu.edu/library/ruby-2.6.6-node-15:20210507 as base
```
awead, line 1: Could go ahead and move to ruby 2.7.3? Or is that moving too many things at once?

Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+
+COPY --chown=app . /app
+
+RUN RAILS_ENV=production \
+  SECRET_KEY_BASE=rails_bogus_key \
+  MYSQL_USER=mysql \
+  MYSQL_PASSWORD=mysql \
+  MYSQL_HOST=mysql \
+  MYSQL_DATABASE=mysql \
+  bundle exec rails assets:precompile && \
```
awead, line 55: I didn't think we were using asset pipeline. There is no `app/assets` directory. I thought everything was in webpacker.

awead (changes requested):
I had two minor questions, but overall this is really starting to come together!

Dockerfile
awead: Nm. It's calling webpack.

Dockerfile
whereismyjetpack: I think we could tackle upgrading ruby as another PR.

awead (approved):

Dockerfile
whereismyjetpack: I was under the impression that assets:precompile was an abstraction to webpacker/sprockets/whatever else and the app just _did the right thing_ this does mirror what we do in prod today, but if we can streamline it i'd be for that, too 

https://github.com/psu-libraries/psulib_blacklight_deploy/blob/d3c8a434f8dcdcd112c425b749a56f3cdbc60050/roles/ansible-rails_app/handlers/main.yml#L10

banukutlu (dismissed):
lgtm 👍🏻  thanks @whereismyjetpack 🚀  :shipit:

Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+ARG UID=2000
+
+USER root
+RUN apt-get update && \
+   apt-get install --no-install-recommends -y \
+   sqlite3 \
+   vim \
+   default-libmysqlclient-dev \
+   libsqlite3-dev \
+   shared-mime-info && \
```
awead, line 11: @whereismyjetpack and I figured out `vim` doesn't need to be there. What about `sqlite3` and the others?

awead (changes requested):

Dockerfile
whereismyjetpack: SQLite needs to be in the dev stage(for now)  thanks for shining a light on this, I think at one point they were in the right stages and some copy pasta happened. 


Mmmm : 🍝

whereismyjetpack: for some reason the bundle won't install without sql dev packages even if we only install the production group. I'd like to, as part of another PR hook up mysql into the docker-compose setup and remove the dep on sqlite so that the local more closely matches prod anyway

awead: that's really weird. We don't have any of those in the scholarsphere Dockerfile, but that's using postgres.

whereismyjetpack: the Gemfile has in it sqlite3, i think bundle downloads gems even if they aren't installed to do the dependency resolution or something, i think eventually they will be pruned out of there. 

"While the --without option will skip installing the gems in the specified groups, it will still download those gems and use them to resolve the dependencies of every gem in your Gemfile(5)."


https://bundler.io/v2.2/man/bundle-install.1.html

awead: what about https://github.com/psu-libraries/scholarsphere/blob/develop/Dockerfile#L41

whereismyjetpack: yeah, i tried that too -- same results so i left it

whereismyjetpack: I'm a liar. i gave it a csv. i will try that

whereismyjetpack: that was it, moved sql* packages down to the dev stage

awead (dismissed):

Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN chown -R app /app
 USER app
 
 COPY Gemfile Gemfile.lock /app/
-RUN gem install bundler:2.1.4
+RUN gem install bundler -v "$(grep -A 1 "BUNDLED WITH" Gemfile.lock | tail -n 1)"
```
awead, line 17: nice

awead:
Looks good. I was able to build it locally and get solr up and running with a custom port.

awead (approved):

PR #773, opened 2021-06-25
Silence deprecation warnings in production

awead:
These are clogging up our logs and only really need to be displayed for testing and development.

### DIFF
--- a/config/initializers/deprecation_silencers.rb
+++ b/config/initializers/deprecation_silencers.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Silence deprecation warnings made in the deprecation gem and in active-support
+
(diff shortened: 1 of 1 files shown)


whereismyjetpack:
Thanks for taking the time to not only look into this, but implement it!

whereismyjetpack (approved):

awead:
We should do the same for Scholarsphere...

whereismyjetpack:
I also didn't look into how critical these are to actually mitigate, don't know if we should ticket it or not...

awead:
If they're done correctly, it means you should fix them before a major version change, but it's not something as critical as a security vulnerability. So I'm completely okay with moving these to dev/test only. That's where they would get directly addressed anyway.

Issue #737, opened 2021-05-06
Change ETAS box text approx 2 weeks out from ETAS EOL


ruthtillman:
About 2 weeks out from the ETAS EOL we'll want to change the descriptive text to state that ETAS access will end on ____ date. We do not yet have a date. This text is defined in `locales/en.yml`, so easy to update.

Current text: 

```
By special arrangement in response to the COVID-19 pandemic, patrons may access this resource online through the HathiTrust Emergency Temporary Access Service.
```

Updated text:

[tbd]

PR #774, opened 2021-06-28
updated ETAS messaging to reflect July 9 changeover

ruthtillman:
Giving folks 2 weeks warning for the ETAS ending #737

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -38,7 +38,7 @@ en:
         public_domain_text: 'Access digital copy through HathiTrust'
         restricted_access_text: 'Search inside at HathiTrust'
         etas_text: 'Check out digital copy through HathiTrust'
-        etas_additional_text: 'By special arrangement in response to the COVID-19 pandemic, patrons may access this resource online through the HathiTrust Emergency Temporary Access Service.'
+        etas_additional_text: 'Digital access to this item through the HathiTrust Emergency Temporary Access Service will conclude on July 9, 2021. Requests for most library materials will be possible through "I Want It." On-shelf access will resume as libraries fully re-open over the summer. We will retain digital access to public domain materials in HathiTrust.'
       facet_tooltips:
         access_facet:
           free_to_read: 'These materials don''t require a Penn State login'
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -38,7 +38,7 @@ en:
         public_domain_text: 'Access digital copy through HathiTrust'
         restricted_access_text: 'Search inside at HathiTrust'
         etas_text: 'Check out digital copy through HathiTrust'
-        etas_additional_text: 'By special arrangement in response to the COVID-19 pandemic, patrons may access this resource online through the HathiTrust Emergency Temporary Access Service.'
+        etas_additional_text: 'Digital access to this item through the HathiTrust Emergency Temporary Access Service will conclude on July 9, 2021. Requests for most library materials will be possible through "I Want It." On-shelf access will resume as libraries fully re-open over the summer. We will retain digital access to public domain materials in HathiTrust.'
```
awead, line 41: A nice yaml trick is to use ">" which allows line breaks. It really makes it easier to read the diffs:

``` yaml
etas_additional_text: >
  Digital access to this item through the HathiTrust Emergency Temporary Access Service will conclude on
  July 9, 2021. Requests for most library materials will be possible through "I Want It." On-shelf access will
  resume as libraries fully re-open over the summer. We will retain digital access to public domain materials
  in HathiTrust.
```
Also, the opening and closing single quotes aren't needed for string values.

ruthtillman: I wondered since we don't in some other places, but since the rest were in single quotes, I erred on the side of not messing w/the pattern.

awead: There are only a handful of cases where you need to use the quotes, and if you're checking the output, you'd notice it right away. But, yeah, consistency is nice.

awead (approved):

spec/views/external_links/_hathi_links.html.erb_spec.rb
```diff
@@ -17,6 +17,6 @@
     expect(rendered).to have_link(href: 'https://catalog.hathitrust.org/Record/12345')
       .and have_css("img[src*='HathiTrust_logo']")
       .and include(I18n.t('blackcat.hathitrust.public_domain_text'))
-      .and include I18n.t('blackcat.hathitrust.etas_additional_text')
+      .and have_css('p.record-view-only', text: I18n.t('blackcat.hathitrust.etas_additional_text'))
```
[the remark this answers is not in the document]
awead: We can ticket this separately.

Issue #780, opened 2021-07-10
Remove HathiTrust box from search results


ruthtillman:
Now that ETAS is gone, it's not a useful part of the search results, even if public domain. Let's just remove it for now and can consider adding in a public domain-only version if we want. There's some later enhancement work to do adding google books stuff so we might evaluate it all there.

nicktiberi:
@ruthtillman - can you point me to a page where this box still appears? I'm not seeing it in prod anymore since ETAS was turned off, but maybe I'm not looking in the right place.

ruthtillman:
https://catalog.libraries.psu.edu/?search_field=all_fields&q=conversations+with+god

ruthtillman:
Looks great!

PR #784, opened 2021-07-13
Remove the HathiTrust box from search results

nicktiberi:
Re: #780.

The HathiTrust box will still be shown in the individual item view.

### DIFF
--- a/app/views/external_links/_all.html.erb
+++ b/app/views/external_links/_all.html.erb
@@ -6,4 +6,4 @@
 
 <%= render partial: 'external_links/access_online_links', locals: { access_online_links: access_online_links } if access_online_links.present? %>
 
-<%= render partial: 'external_links/hathi_links', locals: { hathi_links: document.hathi_links } if hathi_links %>
+<%= render partial: 'external_links/hathi_links', locals: { hathi_links: hathi_links } if hathi_links && show_hathi_links %>
--- a/app/views/catalog/_index_default.html.erb
+++ b/app/views/catalog/_index_default.html.erb
@@ -24,7 +24,11 @@
       <% end -%>
     </ul>
 
-    <%= render partial: 'external_links/all', locals: { document: document } %>
+    <%= render partial: 'external_links/all',
+               locals: {
+                 document: document,
+                 show_hathi_links: document.hathi_links && document.hathi_links[:etas_item]
+               } %>
   </div>
 </div>
 
--- a/app/views/catalog/_show_top_fields.html.erb
+++ b/app/views/catalog/_show_top_fields.html.erb
@@ -13,4 +13,4 @@
   <% end %>
 </dl>
 
-<%= render partial: 'external_links/all', locals: { document: document } %>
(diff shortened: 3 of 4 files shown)


app/views/external_links/_all.html.erb
```diff
@@ -1,9 +1,6 @@
 <% psu_digital_collections_links = document.psu_digital_collections_links
-   access_online_links = document.access_online_links
-   hathi_links = document.hathi_links %>
+   access_online_links = document.access_online_links %>
 
 <%= render partial: 'external_links/psu_digital_collections_links', locals: { psu_digital_collections_links: psu_digital_collections_links } if psu_digital_collections_links.present? %>
 
 <%= render partial: 'external_links/access_online_links', locals: { access_online_links: access_online_links } if access_online_links.present? %>
-
-<%= render partial: 'external_links/hathi_links', locals: { hathi_links: document.hathi_links } if hathi_links %>
```
banukutlu, line 9: @nicktiberi Since we still have the rest of the hathi feature in the code, instead of removing these, can we use the check we used here https://github.com/psu-libraries/psulib_blacklight/blob/main/app/views/catalog/_show_availability.html.erb#L8?

banukutlu (changes requested):

app/views/external_links/_all.html.erb
nicktiberi: @banukutlu - if I add the check for `@document.hathi_links.dig(:etas_item)`, will that fail now for everything because ETAS is over?

If so, then the Hathi box will never appear, and I'm not sure if that's what Ruth wants. My understanding was she wanted to keep the Hathi box for the individual item view, but remove the title, and for search results view, remove the box entirely.

Sorry for any confusion on this!

banukutlu: > @banukutlu - if I add the check for `@document.hathi_links.dig(:etas_item)`, will that fail now for everything because ETAS is over?

Yes, it won't find any etas item when Hathi ETAS set to false.

> 
> If so, then the Hathi box will never I, and I'm not sure if that's what Ruth wants. My understanding was she wanted to keep the Hathi box for the individual item view, but remove the title, and for search results view, remove the box entirely.
> 
> Sorry for any confusion on this!

Your understanding is right, with this change I do not mean to change the rules, just wanted to keep things working same to be able to set back if Hathi ETAS needed to be set back to true.

nicktiberi:
@banukutlu - I made some changes so now the HathiTrust box should reappear in search results if we turn ETAS back on.

banukutlu:
this looks good to me, thanks! wondering if it makes sense/worth it to add a view spec for this? what do you think?

nicktiberi:
> this looks good to me, thanks! @nicktiberi wondering if it makes sense/worth it to add a view spec for this? what do you think?

Yeah for sure! Definitely makes sense to test the show/hide logic. Will add a spec for this.

banukutlu (approved):
👌🏻  thanks Nick!

Issue #781, opened 2021-07-10
HathiTrust box title


ruthtillman:
Title of HathiTrust box is "Full Text Available Online" -- which isn't really true for search inside books. We might need to split it between types or we might need to do something else? I need to review this before it's actually done though.

* [x] Ruth: Review and address titles of HathiTrust ETAS boxes
* [ ] Nick?: Change title

ruthtillman:
Having reviewed this and looked at it, I think we should err on the side of just removing the title as it's extra screen space. Beforehand we only had Public Domain on there AFAIK and thus "Full Text" -- either way we end up presenting HT, it's just extra space we don't need that's already explained in the link text. @nicktiberi ball's in your court now.

ruthtillman:
Looks good. Checked on both Public Domain and Search Inside items!

PR #785, opened 2021-07-13
Remove title from HathiTrust box

nicktiberi:
Re: #781.

### DIFF
--- a/app/views/external_links/_hathi_links.html.erb
+++ b/app/views/external_links/_hathi_links.html.erb
@@ -1,6 +1,10 @@
 <div class="bs-callout bs-callout-warning">
-  <h5 class="record-view-only" id="conveying-meaning-to-assistive-technologies">Full Text available online</h5>
-  <p class="record-view-only"><%= hathi_links[:additional_text] %></p>
+  <% if hathi_links.dig(:etas_item) -%>
+    <h5 class="record-view-only" id="conveying-meaning-to-assistive-technologies">Full Text available online</h5>
+  <% end %>
+  <% if hathi_links[:additional_text] -%>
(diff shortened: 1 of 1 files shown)


app/views/external_links/_hathi_links.html.erb
```diff
@@ -1,5 +1,4 @@
 <div class="bs-callout bs-callout-warning">
-  <h5 class="record-view-only" id="conveying-meaning-to-assistive-technologies">Full Text available online</h5>
```
banukutlu, line 2: @nicktiberi can we use this check https://github.com/psu-libraries/psulib_blacklight/blob/main/app/views/catalog/_show_availability.html.erb#L8 instead of deleting?

banukutlu (changes requested):

nicktiberi:
@banukutlu - with the other PR, #784, the HathiTrust box will no longer appear in search results. That means the only place the HathiTrust box will appear is in the individual item view, right? If so, then I think it makes sense to just delete the text altogether since there is no situation in which it would appear. 

I think what I've done here makes sense, but please let me know if I'm missing something. Thanks! 🙂

banukutlu:
> @banukutlu - with the other PR, #784, the HathiTrust box will no longer appear in search results. That means the only place the HathiTrust box will appear is in the individual item view, right? If so, then I think it makes sense to just delete the text altogether since there is no situation in which it would appear.
> 
> I think what I've done here makes sense, but please let me know if I'm missing something. Thanks! 🙂

I was thinking the situation if we needed to switch back to Hathi ETAS. Yes, currently it won't and should not show these with Hathi ETAS set to false.

banukutlu (approved):
👌🏻

Issue #760, opened 2021-06-14
Refactor Availability 


banukutlu:
- View Components?

PR #786, opened 2021-07-16
Availability: React Refactor Phase 1

nicktiberi:
Re: #760.

@banukutlu - I spent some time this week experimenting with refactoring the availability JS in React. This is not a complete refactor, but I took the main portion of the HTML generating jQuery code and converted them into React components. We could build on this if we decide to continue down this path.

The basic gist of what I did was to create a base `<Availability />` component, which handles rendering the availability/holdings table. Within the `<Availability />` component, there is a `<HoldingDetails />` component with 4 sub-components:
- CourseReserveDueDate
- LocationInfo
- MapScanLink
- PublicNote

The advantage of this is that it allows us to separate out the various bits of logic that control the availability (parsing SIRSI responses, rendering UI, etc.).

My main reason for choosing React over [ViewComponent](https://viewcomponent.org/) was because we are rendering up to 100 of these components on a page at once, and since it is done asynchronously, my understanding was that we would either have to make 100 requests to the Rails endpoint that generates the ViewComponent (sounds v bad), or make a request to some Rails endpoint that generates multiple ViewComponents, and then individually place them on the page (less bad). The React approach seems more straightforward, IMO, but then again this is just my opinion and my natural tendency. I tested the performance with 100 React components in the search results and it seemed totally fine, which was great news!

No rush to check this out, of course, since of course our main priority is call number browse, but I'm curious to see what you think! 🙂 

### Still To Do
- [x] incorporate JSX/React linting

### Maybe Later
- [ ] move even more of the jQuery UI code (showing/hiding buttons, etc) into React components
- [ ] consider adding [Jest](https://github.com/facebook/jest) and unit test this at the individual React component level
- [ ] incorporate TypeScript for better autocomplete/tooling and type safety/error checking

### DIFF
--- a/app/javascript/availability/components/availability.jsx
+++ b/app/javascript/availability/components/availability.jsx
@@ -1,0 +1,56 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import HoldingDetails from './holding_details.jsx';
+import ViewMoreButton from './view_more_button.jsx';
+
+const Availability = ({data}) => { return (
+    <>
+        { data.map((element, index) => {
+            const holdings = element.holdings;
+            const catkey = holdings[0].catkey;
+            const uniqueID = catkey + index;
+            const moreHoldings = holdings.length > 4 ? holdings.splice(4, holdings.length) : [];
+
+            return (
+                <div key={index} data-library={element.summary.libraryID}>
+                    <h5>{element.summary.library} ({element.summary.countAtLibrary} {element.summary.pluralize})</h5>
+
+                    <table id={"holdings-" + uniqueID} className="table table-sm">
+                        <caption className="sr-only">Listing where to find this item in our buildings.</caption>
+                        <thead className="thead-light">
+                            <tr>
+                                <th>Call number</th>
+                                <th>Material</th>
+                                <th>Location</th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                            { holdings.map((holding, index) => (
+                                <tr key={index}>
+                                    <HoldingDetails holding={holding} />
+                                </tr>
+                            )) }
+
+                            { moreHoldings.map((holding, index) => (
+                                <tr key={index} className="collapse" id={"collapseHoldings" + uniqueID}>
+                                    <HoldingDetails holding={holding} />
+                                </tr>
+                            )) }
+                        </tbody>
+                    </table>
+
+                    { (moreHoldings.length > 0) && (
+                        <ViewMoreButton uniqueID={uniqueID} />
+                    ) }
+                </div>
+            )
+        }) }
+    </>
+)};
+
+// eslint-react: defines valid prop types passed to this component
+Availability.propTypes = {
+    data: PropTypes.array
+};
+
+export default Availability;
--- a/app/javascript/availability/components/course_reserve_due_date.jsx
+++ b/app/javascript/availability/components/course_reserve_due_date.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import availability from '../index.js';
+
+const CourseReserveDueDate = ({holding}) => {
+    if (availability.isOnCourseReserve(holding)) {
+        const dueDate = new Date(holding.dueDate);
+
+        const day = dueDate.toLocaleDateString('en-us', {
+            year: 'numeric',
+            month: 'numeric',
+            day: 'numeric',
+            timeZone: 'America/New_York'
+        });
+
+        const time = dueDate.toLocaleTimeString('en-us', {
+            hour: 'numeric',
+            minute: '2-digit',
+            timeZone: 'America/New_York'
+        });
+
+        const circulationRule = availability.reserveCirculationRules[holding.reserveCirculationRule];
+
+        return (
+            <div>
+                <strong>Due back at:</strong> {time} on {day}<br />{circulationRule}
+            </div>
+        );
+    }
+
+    return null;
+};
+
+// eslint-react: defines valid prop types passed to this component
+CourseReserveDueDate.propTypes = {
+    holding: PropTypes.object
+};
+
+export default CourseReserveDueDate;
--- a/app/javascript/availability/components/holding_details.jsx
+++ b/app/javascript/availability/components/holding_details.jsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import PublicNote from './public_note.jsx';
+import LocationInfo from './location_info.jsx';
+import CourseReserveDueDate from './course_reserve_due_date.jsx';
+import MapScanLink from './map_scan_link.jsx';
+
+const HoldingDetails = ({holding}) => {
+    const generateCallNumber = () => {
+        // Do not display call number for on loan items
+        return (holding.locationID === "ILLEND") ? "" : holding.callNumber;
+    };
+
+    return (
+        <>
+            <td>
+                {generateCallNumber(holding)}
+                <PublicNote holding={holding} />
+            </td>
+            <td>{holding.itemType}</td>
+            <td>
+                <LocationInfo holding={holding} />
+                <CourseReserveDueDate holding={holding} />
+                <MapScanLink holding={holding} />
+            </td>
+        </>
+    );
+};
+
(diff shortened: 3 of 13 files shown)


nicktiberi:
@banukutlu - I added some additional React-specific linting and took care of a few issues that it complained about. We can add more if needed: https://github.com/yannickcr/eslint-plugin-react#list-of-supported-rules

Code Climate is complaining about a few things, but I'm not sure if we care about all of its complaints. I will look into the complaints in `location_info.jsx` - that can probably be simplified a bit more.

app/javascript/availability/components/course_reserve_due_date.jsx
```diff
@@ -0,0 +1,40 @@
...
+    const holding = props.holding;
+
+    if (availability.isOnCourseReserve(holding)) {
+        const dueDate = new Date(holding.dueDate);
+
+        const day = dueDate.toLocaleDateString('en-us', {
+            year: 'numeric',
+            month: 'numeric',
+            day: 'numeric',
+            timeZone: 'America/New_York'
```
awead, line 15: Do we know if `holding.dueDate` is in GMT?

app/javascript/availability/components/location_info.jsx
```diff
@@ -0,0 +1,95 @@
...
+                data-item-type={holding.itemTypeID}
+                data-item-id={holding.itemID}
+                data-item-location={specialCollectionsText}
+                href="#">
+                    <Spinner />
+                    Request Material
+            </a>
+        </>);
+    }
+
```
awead, line 86: I wonder if you can take this one step further and so something like:

``` js
if (availability.isIllLink(holding)) {
  return (
    <IllLink holding={holding} />
  )
}

if (availability.isArchivalThesis(holding)) {
  return (
    <IllLink holding={holding} />
    <AeonLink holding={holding} />
  )
  
if (availability.isArchivalMaterial(holding)) {
return (
    <AeonLink holding={holding} />
  )
}
```

app/javascript/availability/components/location_info.jsx
```diff
@@ -0,0 +1,95 @@
...
+    // Otherwise use the default translation map for location display, no link
+    return availability.mapLocation(availability.allLocations, holding.locationID);
```
awead, line 88: Your second argument is the same as "self", i.e. `availability`. Wouldn't the `mapLocation` function be able to call `allLocations` from within itself?

awead (changes requested):
Overall this looks good. I like seeing the re-useablility of the different React components, and I think it's an improvement. I had a couple of suggestions, but keep in mind I don't React, so some of my comments may be completely off.

I do see a lot of coupling of availability with a holding. If it was looking at this from a Ruby perspective, I might think about an Availability class, or component, has having a holding with it, since it seems that most of the time we're having to pass a holding into one of the availability's functions. That's just an observation, and not a suggestion, so take that as you want.

app/javascript/availability/components/course_reserve_due_date.jsx
nicktiberi: I believe it is, but I am not 100% certain. I had to add the `timeZone` param in a previous commit because the CI server was running in a different time zone than the catalog server, which was causing tests to fail on CI.

app/javascript/availability/components/location_info.jsx
nicktiberi: @awead - good call. I was thinking of saving that for a second stage of refactor but maybe it makes sense to just do this now. There is some more ILL/AEON logic in `availability/index.js` that can then be moved into these components in a later stage of the refactor.

app/javascript/availability/components/location_info.jsx
nicktiberi: Ha! Yeah, this was just a copy paste, and it seems the first argument to `mapLocation` is always `allLocations`. Will change that. Good catch!

nicktiberi:
> Overall this looks good. I like seeing the re-useablility of the different React components, and I think it's an improvement. I had a couple of suggestions, but keep in mind I don't React, so some of my comments may be completely off.
> 
> I do see a lot of coupling of availability with a holding. If it was looking at this from a Ruby perspective, I might think about an Availability class, or component, has having a holding with it, since it seems that most of the time we're having to pass a holding into one of the availability's functions. That's just an observation, and not a suggestion, so take that as you want.

Yeah, that is a good point. We may be able to move some more functions out of the mega `availability/index.js` file into these components in a later stage of the refactor and eliminate some more of these `availability.doSomething(holding)` calls.

app/javascript/availability/components/location_info.jsx
awead: Up to you. If you want to "tie it off" now and save it for later, that's ok by me.

app/javascript/availability/components/location_info.jsx
nicktiberi: @awead - moved the `mapLocation` function into this component and also removed the unnecessary argument. Whittling away at the mega file little by little. 🙂 

If you and @banukutlu are ok with it, I'd like to handle the AEON/ILL stuff in a separate PR, since this one is getting kinda big already.

awead (approved):

nicktiberi:
@banukutlu - this should be ready to go, but I'll wait for your stamp of approval. 🙂

banukutlu:
> @banukutlu - this should be ready to go, but I'll wait for your stamp of approval. 🙂

cool 😎 , I will get to it after the LC review, hopefully sometime tomorrow :)

app/javascript/availability/components/location_info.jsx
banukutlu: I was exactly thinking the same thing like it would be very nice to see `<IllLink holding={holding} />` and `<AeonLink holding={holding} />` components. I am totally fine for these to be the next phase of the refactor since as Nick mentioned this PR has already good amount of changes.

app/javascript/availability/components/holding_details.jsx
```diff
@@ -0,0 +1,33 @@
...
+            <td>
+                <LocationInfo holding={holding} />
+                <CourseReserveDueDate holding={holding} />
+                <MapScanLink holding={holding} />
+            </td>
+        </>
+    );
+};
+
+// eslint-react: defines valid prop types passed to this component
```
banukutlu, line 28: thanks @nicktiberi  for adding these :)

app/javascript/availability/components/availability.jsx
```diff
@@ -0,0 +1,61 @@
...
+                            { moreHoldings.map((moreHolding, index) => (
+                                <tr key={index} className="collapse" id={"collapseHoldings" + uniqueID}>
+                                    <HoldingDetails holding={moreHolding} />
+                                </tr>
+                            )) }
+                        </tbody>
+                    </table>
+
+                    { (moreHoldings.length > 0) ? (
+                        <b><button className="btn btn-primary toggle-more"
```
banukutlu, line 42: @nicktiberi would it worth it to make toggle more as a component?

nicktiberi: @banukutlu - added a `<ViewMoreButton />` component and did a few more small cleanups with passing props into components, so this should be ready to go! 😎

app/javascript/availability/components/view_more_button.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
banukutlu, line 1: thanks @nicktiberi :)

app/javascript/availability/components/availability.jsx
```diff
@@ -0,0 +1,56 @@
...
+                            { moreHoldings.map((holding, index) => (
+                                <tr key={index} className="collapse" id={"collapseHoldings" + uniqueID}>
+                                    <HoldingDetails holding={holding} />
+                                </tr>
+                            )) }
+                        </tbody>
+                    </table>
+
+                    { (moreHoldings.length > 0) && (
+                        <ViewMoreButton uniqueID={uniqueID} />
```
banukutlu, line 43: 👌🏻

app/javascript/availability/components/availability.jsx
banukutlu: awesome, thanks 😎

Issue #787, opened 2021-07-19
Map scan links not showing in "View More" expansion


ruthtillman:
The scan links are showing on the first 4, but when you expand to "View More," they aren't loading. Suspect it's how we're doing the JavaScript on this. Tagging Nick to investigate.

https://catalog.libraries.psu.edu/catalog/1677791

ruthtillman:
Perfect! Thanks for the quick fix.

PR #788, opened 2021-07-19
Show map scan links and public notes for holdings behind "View More"

nicktiberi:
Re: #787.

Holdings hidden behind the "View More" button were not showing map scan links or public note tooltips because the code to render them wasn't included in that block.

We could consider combining the holdings/moreHoldings blocks into a single reusable component to reduce code duplication and bugs like this one.

### Before
Screen Shot 2021-07-19 at 10 01 56 AM

### After
Screen Shot 2021-07-19 at 10 02 06 AM


banukutlu (approved):
✔️

banukutlu (changes requested):
@nicktiberi can we add a test for this?

banukutlu (approved):
lgtm 👌🏻  thanks!

banukutlu:
> We could consider combining the holdings/moreHoldings blocks into
a single reusable component to reduce code duplication and bugs
like this one.

yup ☝🏻

PR #792, opened 2021-07-22
Libraries locations

ruthtillman:
Updated both libraries and locations information to handle request from CCLs + also to handle recent locations changes. (locations in Atom appear to have very slightly changed tabbing so it's looking like a full find/replace vs. a wipe ... I did diff both files on my own to review that part).


ruthtillman:
Btw I think the reason it's a larger change is because of how the replacement on Locations was a batch replacement that happened to tab differently so it becomes a full replacement not a partial replacement of a big section.

app/javascript/availability/libraries_locations.json
```diff
@@ -756,15 +765,15 @@
...
     "WORLD": "World Campus",
-    "WSCRANTON": "Scranton",
+    "WSCRANTON": "Penn State Scranton",
     "XTERNAL": "Associated Libraries",
-    "YORK": "York"
+    "YORK": "Penn State York"
   },
   "request_via_ill": {
-    "BINDERY-DN": "Copy unavailable, request via Interlibrary Loan",
+    "BINDERY-DN": "Copy unavailable, request via Interlibrary Loan".
```
nicktiberi, line 776: @ruthtillman - looks like you have a period here in the JSON instead of a comma, which is a syntax error. This may be what is causing a bunch of tests to fail. I bet if you swapped this back out CI will be green again 🟢  🙂

ruthtillman: oh good grief. Thanks so much for the review.

banukutlu (approved):
👌🏻

PR #799, opened 2021-07-30
Call number browse UI

nicktiberi:
### Exact Match
Screen Shot 2021-08-02 at 3 30 20 PM

### Nearby
Screen Shot 2021-08-02 at 3 30 35 PM

### DIFF
--- a/spec/views/browse/index.html.erb_spec.rb
+++ b/spec/views/browse/index.html.erb_spec.rb
@@ -1,0 +1,134 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'browse/index', type: :view do
+  let(:shelf_list) { ShelfListPresenter.new }
+
+  let(:doc1) {
+    SolrDocument.new({
+                       id: '123',
+                       call_number_ssm: ['abc'],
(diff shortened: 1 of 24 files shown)


spec/views/browse/index.html.erb_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+      expect(rendered).to have_selector 'tr:nth-child(3) td:nth-child(2)',
+                                        exact_text: 'book 3'
+      expect(rendered).to have_selector 'tr:nth-child(3) td:nth-child(3)',
+                                        exact_text: 'library 3'
+      expect(rendered).to have_selector 'tr:nth-child(3) td:nth-child(4)',
+                                        exact_text: 'pub info 3'
+    end
+  end
+
+  # TODO: implement these tests properly
```
nicktiberi, line 80: @awead - maybe we can talk about this on Monday. I am not sure if I'm fully understanding how this list is being built and want to make sure I'm testing this properly. Right now these tests are super incomplete and/or incorrect.

nicktiberi: Per convo with @awead, we can worry about cleaning up these tests later.

Issue #793, opened 2021-07-23
LC Call Number Browse Display


banukutlu:
Basic call number browse, with a simple interface similar to Princeton's [implementation](https://catalog.princeton.edu/browse/call_numbers?start=3800).

Note: This will _only_ cover LC-based call number classifications.

## TODO

- [ ] Sort and Search Call Numbers 
   - no search param: start from beginning with pagination
   - search for an existing call number: display the item with N before and after items
   - search for non existing/partial call number: display the non existing item in the context of related items
- [ ] Display Call Numbers
   - List of base call numbers with title(`title_display_ssm`) with link to the record, author, publication date, and image (same thumbnail used for item display) for the kind of bookshelf browse view
    
## Questions

1. Multiple solr documents(records) with same base call number

See https://github.com/psu-libraries/psulib_blacklight/wiki/Browse-Items-by-Library-of-Congress-Call-Number

awead:
## Second Sprint TODO

* use shelf keys to browse forward and backward (and not the raw call numbers)

## Questions

How do we want to display multiple call numbers? Ex:

|Call Numnber | Title  | Location | Publication Info |
--- | --- | --- | ---- |
| AB 1234.5 2002 | Title 1 | Here | Info 1 |
|  | Title 2 | There | Info 2 |
|  | Title 3 | There | Info 3 |

### OR

|Call Numnber | Title  | Location | Publication Info |
--- | --- | --- | ---- |
| AB 1234.5 2002 | Title 1 | Here | Info 1 |
| AB 1234.5 2002 | Title 2 | There | Info 2 |
| AB 1234.5 2002 | Title 3 | There | Info 3 |

### OR 

?????

banukutlu:
> ## Questions
> How do we want to display multiple call numbers? Ex:
> 
> ### OR
> Call Numnber	Title	Location	Publication Info
> AB 1234.5 2002	Title 1	Here	Info 1
> AB 1234.5 2002	Title 2	There	Info 2
> AB 1234.5 2002	Title 3	There	Info 3

Above is how we want to display multiple call numbers.

ruthtillman:
Reviewed what we've got so far and it's going VERY well! I documented 2 issues, one in Blacklight display (#831) and one which captures small regex-related things https://github.com/psu-libraries/psulib_traject/issues/343. 

Overall I really enjoyed browsing through and seeing how well it works and actually getting to view all our books on quilting in one place (I also tested other sections of the call numbers)

PR #807, opened 2021-08-04
Shelf lists, holdings, and items for browsing call numbers

awead:
Adds a new feature enabling LC call number browsing. A ShelfList service performs a Solr term component query to retrieve an ordered list of call numbers before and after a given call number query. This could be an exact or partial LC number.

A ShelfListPresenter assembles the service's results into a listing of ShelfItem objects, and orders them as they would appear on a virtual "shelf".  The user can browse forward or backward from a given point, or search to see which items are near a given number.

Multiple records may occur within a single shelf item because they may share the same base call number. Likewise, the same catalog record can occur multiple times within a shelf because it may have more than one different call number.

## Traject Dependency

This PR depends on a change in traject https://github.com/psu-libraries/psulib_traject/pull/333. The `keymap_struct` is updated so that the catalog can more easily determine which key is associated with which call number.

Related to #793


app/services/shelf_list.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    documents = document_query(field: shelfkey, values: keys)
+    holdings = Holdings.new(documents: documents, direction: direction)
+    keys.map do |key|
+      holdings.find(key)
+    end
+  end
+
+  # @note The first Solr query used to get a listing of forward or reverse shelf keys for a given term query.
+  def terms_query(field:, limit:, value: nil)
```
banukutlu, line 70: Can this be a private method?

app/services/shelf_list.rb
```diff
@@ -0,0 +1,104 @@
...
+           })
+      .dig('terms', field) || []
+  end
+
+  # @note This is the second Solr query, performed after we've retrieved a listing of forward or reverse shelf keys.
+  # Using a specific 'shelf' search handler, we retrieve all the documents for a given set of shelf keys, with a
+  # pre-defined set of fields. Note that the max rows is set in the handler, so it is theoretically possible to NOT get
+  # all of the records back for a given shelf key. But this seems unlikely as the shelf key would need to occur in more
+  # documents than the max set in the handler.
+  def document_query(field:, values:)
```
banukutlu, line 93: Can this be a private method?

app/services/shelf_list.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def reverse_keys
+    @reverse_keys ||= terms_query(
+      field: reverse_shelfkey,
+      value: ShelfKey.reverse(query),
+      limit: reverse_limit
+    )
+      .each_slice(2)
+      .map(&:first)
+  end
```
banukutlu, line 52: can these two methods be private as well? 🤔

app/services/shelf_list.rb
```diff
@@ -0,0 +1,106 @@
...
+        .each_slice(2)
+        .map(&:first)
+    end
+
+    # @return Array<ShelfItem>
+    # @note Uses a Holdings object to build a set of shelf item objects from #document_query. A set of documents is
+    # retrieved from Solr for a given set of shelf keys. The keys are already in the proper order, we just need a
+    # complete document for each one. Using Holdings, we construct a hash of every shelf key and its document from
+    # #document_query.  There can be multiple documents per key, if the same call number is present in multiple items.
+    def holdings(keys:, shelfkey:, direction:)
```
banukutlu, line 61: @awead I kinda wanna rename this method since there is the Holdings object, but I do not have a better idea than holdings :)

spec/services/shelf_list_spec.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ShelfList do
+  describe '#documents' do
+    # These tests are based off of our existing sample data. If our fixture data changes, these tests may break as a
+    # result, requiring them to be updated, but it should not indicate that the ShelfList class itself is at fault.
+    context 'when querying with shelf key that is present in the library' do
```
banukutlu, line 9: @awead thanks for these specs! very helpful for documentation as well 👌🏻

app/services/shelf_list.rb
awead: `shelf_items`?

banukutlu: yeah that is exactly what I thought as an option too. So yeah I kinda like that, if you are OK with it too

awead: @banukutlu all fixed up. This was a good catch. I was already using `holdings` as a variable internal to the method, which is bad style.

banukutlu: awesome, thanks 😎

app/presenters/shelf_list_presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+      elsif ending.present?
+        { query: ending, forward_limit: 1, reverse_limit: length + 1 }
+      elsif nearby.present?
+        { query: nearby, forward_limit: (length - MIN) + 2, reverse_limit: MIN }
+      else
+        { query: '0', forward_limit: length + 1, reverse_limit: 0 }
+      end
+    end
+
+    def beginning?
```
banukutlu, line 114: @awead does it make sense to rename this one as `the_start` 🤔

awead: It's returning true/false. I initially had something like `at_the_beginning?` but that was too long.

banukutlu: `the_start?` because `the_end?` is what I was thinking, just forgot to add the `?`  to them in my comment.

banukutlu: if you think it sounds weird, it is OK no need to change.

app/presenters/shelf_list_presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    # @note There are four possible scenarios, each determined by the parameters given to the presenter, and each
+    # resulting in a specific set of arguments passed to the ShelfList service.
+    #
+    # 1) Querying from a known point forward:
+    #    Returns the number of items after a given key. The two items returned before the key are the item itself, and
+    #    the previous item which is used to construct the link to the previous shelf.
+    # 2) Querying from a known point backwards:
+    #    Returns the number of items before a given key. The one item that occurs after the key is used to construct the
```
banukutlu, line 95:
```suggestion
    #    Returns a set of items before a given key. The one item that occurs after the key is used to construct the
```

app/presenters/shelf_list_presenter.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    def shelf_list
+      @shelf_list ||= ShelfList.call(shelf_list_params)
+    end
+
+    # @note There are four possible scenarios, each determined by the parameters given to the presenter, and each
+    # resulting in a specific set of arguments passed to the ShelfList service.
+    #
+    # 1) Querying from a known point forward:
+    #    Returns the number of items after a given key. The two items returned before the key are the item itself, and
```
banukutlu, line 92:
```suggestion
    #    Returns a set of items after a given key. The two items returned before the key are the item itself, and
```

banukutlu (changes requested):

app/presenters/shelf_list_presenter.rb
awead: How about first?/last?

awead:
@banukutlu I pushed up some new changes that address beginning vs. the_end and the notes in the comments.

app/presenters/shelf_list_presenter.rb
banukutlu: yes, loved it! more clear and distinct 👌🏻

banukutlu (approved):
awesome work! thanks @awead

PR #808, opened 2021-08-05
only create alias if we're creating the inital collection

whereismyjetpack:
We could get ourselves in a pickle if we were testing a collection, and then restart the pod. (the pod would change the pointer to the last_incremented_collection)

Here, we only create the alias when the collection has been initialized, else we rely on someone, or something to run the create_alias task after verification of the new collection has been completed.

### DIFF
--- a/lib/psulib_blacklight/solr_manager.rb
+++ b/lib/psulib_blacklight/solr_manager.rb
@@ -59,8 +59,13 @@ module PsulibBlacklight
       if next_collection_version == 1
         puts "creating collection #{next_collection_version}"
         create_collection
+        @update_config ||= modify_collection
+        create_alias
       else
         puts 'collection exists skipping.'
+        @update_config ||= modify_collection
+        # Return nil to indicate we didn't create a collection
+        nil
       end
     end
 
--- a/lib/psulib_blacklight/solr_manager.rb
+++ b/lib/psulib_blacklight/solr_manager.rb
@@ -62,10 +62,8 @@ module PsulibBlacklight
         @update_config ||= modify_collection
         create_alias
       else
-        puts 'collection exists skipping.'
         @update_config ||= modify_collection
-        # Return nil to indicate we didn't create a collection
-        nil
+        puts 'collection exists skipping.'
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/psulib_blacklight/solr_manager.rb
```diff
@@ -59,8 +59,13 @@ def initialize_collection
       if next_collection_version == 1
```
awead, line 59: would the next version ever be greater than 1?

lib/psulib_blacklight/solr_manager.rb
```diff
@@ -59,8 +59,13 @@ def initialize_collection
...
         puts "creating collection #{next_collection_version}"
         create_collection
+        @update_config ||= modify_collection
+        create_alias
       else
         puts 'collection exists skipping.'
+        @update_config ||= modify_collection
+        # Return nil to indicate we didn't create a collection
+        nil
```
awead, line 68: It doesn't look like you need to rely on nil here. move the puts to the end.

awead (changes requested):

lib/psulib_blacklight/solr_manager.rb
whereismyjetpack: nil was put there specifically to allow the tests to pass, there's a test that expects nil, and couldn't determine what to put in there. _shrug_ didn't think it was such a big deal, sorry

awead: Explicitly returning nil can sometimes mean something's not quite right. Maybe the tests just need to be updated. I only see the method getting invoked from a rake task, so it would make sense if it was just:

``` ruby
else
  @update_config ||= modify_collection
  puts 'collection exists skipping.'
end
```

lib/psulib_blacklight/solr_manager.rb
whereismyjetpack: yeah, since this gets called at startup, it could be anything

lib/psulib_blacklight/solr_manager.rb
whereismyjetpack: gotcha, could have sworn this failed before, but could have been some of the other changes i needed to make, too.

awead (approved):

PR #814, opened 2021-08-11
Call number browse UI

nicktiberi:
Adds a table view for displaying a paginated list of items sorted by call number, with links to individual items.

### Nearby
Screen Shot 2021-08-11 at 3 23 55 PM

### Exact Match
Screen Shot 2021-08-11 at 3 23 44 PM


ruthtillman:
Love how it's looking! (this is a data thing I've discussed w/Adam so not a worry here -- but if there's more than 3 libraries we'll just say "Multiple Locations")

nicktiberi:
> Love how it's looking! (this is a data thing I've discussed w/Adam so not a worry here -- but if there's more than 3 libraries we'll just say "Multiple Locations")

@ruthtillman - thanks for the feedback! I should be able to make that change in this PR.

app/views/browse/index.html.erb
```diff
@@ -0,0 +1,57 @@
...
+      <th scope="col">Call Number</th>
+      <th scope="col">Title</th>
+      <th scope="col">Location</th>
+      <th scope="col">Publication Info</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @shelf_list.list.map do |shelf_item| %>
+        <% if shelf_item.nearby? && !shelf_item.match? %>
+          <tr class="<%= shelf_item.nearby? ? 'table-primary' : '' %>">
```
awead, line 29: class here will always be `table-primary`

awead (changes requested):
Looks good! I forget to tell you to use `key` instead of the actual call number for the ending and starting links.

app/views/browse/index.html.erb
nicktiberi: Classic copy paste mistake... 🤦

awead: happens to us all...

nicktiberi:
@ruthtillman - I've made the update to show "Multiple Locations" if there are more than 3 locations. Screenshots have been updated above.

@awead - I think all of your change requests have been addressed aside from the path helper comment. Thanks for reviewing!

app/views/browse/index.html.erb
```diff
@@ -0,0 +1,57 @@
...
+            <td colspan="4"><%= shelf_item.call_number %></td>
+          </tr>
+        <% else %>
+          <% shelf_item.documents.map do |doc| %>
+            <tr class="<%= shelf_item.nearby? ? 'table-primary' : '' %>">
+              <td>
+                <a href="/catalog/<%= doc[:id] %>"><%= shelf_item.call_number %></a>
```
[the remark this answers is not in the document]
awead: bah. sigh. well, we'll worry about it later. No point in holding the PR up.

awead (approved):

banukutlu:
seems like this one needs a rebase

awead:
@banukutlu I've rebased this off of the browse branch you just merged, and it's ready for a review from you.

app/models/solr_document.rb
```diff
@@ -38,4 +38,12 @@ class SolrDocument
   # and Blacklight::Document::SemanticFields#to_semantic_values
   # Recommendation: Use field names from Dublin Core
   use_extension(Blacklight::Document::DublinCore)
+
+  def location_display
```
banukutlu, line 42: @nicktiberi @awead What do you think for moving this logic a component since It is mostly about the display? Also kinda thinking that it might be better to keep solr_document as lean as possible for blacklight updates 🤔

banukutlu (commented):
@nicktiberi @awead Is it worth adding a controller spec for browse_controller?

app/models/solr_document.rb
awead: That's an excellent point. ShelfItem can contain many SolrDocument objects, but maybe this an inappropriate use of SolrDocument. We could define a mini-presenter object, ex. `ShelfItem::Record`, serves as the thing that holds the solr document data, and put all the necessary display logic in there. It also opens up the question (in my mind) that maybe ShelfItem is also a presenter and not a model. 🤷 Anyway, @banukutlu's point is well-taken, and this logic ought to be someplace else. Happy to pair up with anyone to work this out.

app/models/shelf_item.rb
```diff
@@ -30,7 +30,17 @@ def nearby?
   end
 
   def add(document)
-    @documents << SolrDocument.new(document)
+    @documents << DecoratedDocument.new(SolrDocument.new(document))
+  end
+
+  class DecoratedDocument < SimpleDelegator
```
banukutlu, line 36: 👌🏻  like this decorator approach

app/components/browse/page_size_selector.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <%= link_to '10',
+                link_path(10),
+                class: 'dropdown-item' %>
+    <%= link_to '20',
+                link_path(20),
+                class: 'dropdown-item' %>
+    <%= link_to '50',
+                link_path(50),
+                class: 'dropdown-item' %>
+    <%= link_to '100',
```
banukutlu, line 21: we might even not let for 100?

app/views/browse/index.html.erb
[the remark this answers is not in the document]
banukutlu: yeah I was about to suggest the link_to as well :)

app/components/browse/page_size_selector.html.erb
ruthtillman: I would be ok with that.

app/views/browse/index.html.erb
```diff
@@ -0,0 +1,57 @@
...
+<div class="row my-3">
+  <div class="col-6">
+    <%= render Browse::PageNavigation.new(prev_item: @shelf_list.previous,
+                                          next_item: @shelf_list.next,
+                                          length: params[:length]) %>
+  </div>
+  <div class="col-6 text-right">
+    <%= render Browse::PageSizeSelector.new(params: params) %>
+  </div>
+</div>
```
banukutlu, line 15: @nicktiberi can we move lines 6-15 to a prev_next_component and render both here and at the bottom instead lines 48-57?

@awead

banukutlu (changes requested):

app/components/browse/page_size_selector.html.erb
banukutlu: @nicktiberi can we remove 100 option then?

spec/features/browse_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+                                      exact_text: 'Multiple Locations'
+
+        expect(page).to have_selector 'tr.table-primary:nth-child(2) td:nth-child(4)',
+                                      exact_text: 'Lincoln : University of Nebraska Press, ©1983-©2001.'
+      end
+    end
+
+    context 'when there is not an exact match' do
+      specify do
+        visit '/browse?nearby=LOL'
```
banukutlu, line 49: 😆

app/views/browse/index.html.erb
awead: One of is a page navigation component, and the other is a page size component. They are similar, but slightly different. Are you suggesting a third component that contains the two of them?

banukutlu: Yeah, I was thinking of a third component that contains all the repeated lines. Ah, right just realized one of them is the size selector but regardless yeah just like doing a one render rather than two with the divs around it moved into a component as well? 🤔

nicktiberi: @banukutlu - makes sense to me! I'm finishing up a wrapper component and will commit in a bit here.

banukutlu: kewl, thanks!!

nicktiberi:
@banukutlu - should be good to go now. @awead saved the day after I blew away his changes. 💥

app/views/browse/index.html.erb
```diff
@@ -0,0 +1,43 @@
+
+<h1>Browse by Call Number</h1>
+
+<h2 class="h4"><%= @shelf_list.list.first.call_number %> to <%= @shelf_list.list.last.call_number %></h2>
+
+<%= render Browse::BrowseControls.new(prev_item: @shelf_list.previous,
```
banukutlu, line 6: thanks! looks neat :)

banukutlu (approved):
🎉  looks great! thanks so much @nicktiberi and @awead!!

PR #818, opened 2021-08-12
Availability: React Refactor Phase 2

### LINKED ISSUE #806: Availability React Refactor Phase 2
- [ ] refactor `availabilitySnippet()` into a React component
- [ ] move `isSomething(holding)` methods into React components
- [ ] create `<IllLink />` and `<AeonLink />` components
- [ ] move `createAeonUrl()` and `createIllUrl()` into the `<IllLink />` and `<AeonLink />` components

nicktiberi:
Closes #806.

Continues the refactor effort by moving more functionality into React components, including:

- AeonLink
- IllLink
- Spinner
- Snippet

Also modifies the Babel and ESLint configs so we no longer have to import React explicitly.


app/javascript/availability/components/map_scan_link.jsx
```diff
@@ -1,17 +1,20 @@
...
-                <a target="_blank" rel="noreferrer" href={availability.mapScanUrl}>Request scan</a>
-            </div>
-        );
+    const isUPSpecialMap = () => {
+        return ['MAPSPEC'].includes(holding.itemTypeID) && ['UP-MAPS'].includes(holding.libraryID);
+    };
+
+    if (!isUPSpecialMap()) {
+        return null;
     }
```
banukutlu, line 12: @nicktiberi I liked this 👌🏻 nicer and more readable than the previous.

app/javascript/availability/components/snippet.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+
+const Snippet = ({data}) => {
+    const librariesText = () => {
```
banukutlu, line 4:
```
   const isSnippet = () => {
        ...
        return ...
    };

    if (!isSnippet()) {
        return null;
    }
```

@nicktiberi Do you think it makes sense to have this one with the same approach you had for some of the others?

nicktiberi: @banukutlu - I changed this logic a bit so that it returns early if there are no copies available.

PR #819, opened 2021-08-13
Install rack-mini-profiler

awead:
This is a very useful tool when trying to determine where bottlenecks are occurring in your page load. I've been using it to tweak Solr queries in the call number browse process.

## Collapsed

Screen Shot 2021-08-13 at 9 15 39 AM

## Expanded

Screen Shot 2021-08-13 at 9 16 21 AM

It's also got some other great features, but this is the most often used one.


app/controllers/application_controller.rb
```diff
@@ -13,4 +13,9 @@ class ApplicationController < ActionController::Base
   ActionController::Parameters.permit_all_parameters = true
 
   helper_method :blackcat_config
+
+  before_action do
+    # @todo Need to limit this to admin users only
+    Rack::MiniProfiler.authorize_request
```
awead, line 19: @whereismyjetpack How would we limit this to our "admin" users? Do we have access to group information for the current user?

whereismyjetpack: For blacklight, we don't. We'd have to hook into devise to make decisions at this layer. We can protect a path at the ingress layer, i.e  `/profilier` let me do some more research

awead: We don't necessarily have to use Devise. We could just do it via session information. When a new session is started, it checks the request for the omniauth information. Is that available currently?

awead: I'm opting for an ENV variable to enable this only in dev and preview instances until we can implement user-based restrictions.

whereismyjetpack (approved):

awead:
@whereismyjetpack thanks!

PR #825, opened 2021-08-18
Update Absorbs fields labels

ruthtillman:
Pretty simple typo fix in catalog controller. Neither label nor field appears to be referred to directly in spec so nothing to change there?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -244,10 +244,10 @@ class CatalogController < ApplicationController
     config.add_show_field 'serials_continues_in_part_display_ssim', label: 'Continues in Part', helper_method: :title_links
     config.add_show_field 'serials_continued_in_part_by_display_ssim', label: 'Continued in Part By', helper_method: :title_links
     config.add_show_field 'serials_formed_from_display_ssim', label: 'Formed From', helper_method: :title_links
-    config.add_show_field 'serials_absorbs_display_ssim', label: 'Asborbs', helper_method: :title_links
-    config.add_show_field 'serials_absorbed_by_display_ssim', label: 'Asborbed By', helper_method: :title_links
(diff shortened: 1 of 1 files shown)


awead:
Is there a test to see if the referenced solr fields are showing up in the record display  somewhere?

ruthtillman:
catkey 34332311 is one I know has them.

banukutlu:
> Is there a test to see if the referenced solr fields are showing up in the record display somewhere?

I do not think so, we were only adding specs for the customization we added. Since there are many fields that just uses out of the box `config.add_show_field`, we did not add specs for them. There is one for the helper method title_links but it does not specifically checks for this field.

banukutlu (approved):

awead:
@banukutlu My only concern is that if these lines get remove/changed, then no tests would break. Does that mean we'd be introducing a regression bug?

banukutlu:
> @banukutlu My only concern is that if these lines get remove/changed, then no tests would break. Does that mean we'd be introducing a regression bug?

correct, no tests covers that, we can consider adding specs. there were many straight forward fields like this so we were not sure about the worth of adding them with thinking if somebody is deleting them mistakenly, hopefully the reviewer would catch on that.

ruthtillman:
Yeah there are a ton of these small fields. It would definitely be something the reviewer should catch. Also people will _let me know_ 😹 

I support writing specs but that might need to happen a little later so we can get a thorough coverage expectation because there are many of these little fields.

awead:
@ruthtillman If you want, go ahead and put in a ticket and describes what we need to test. I could be something as simple as testing the Blacklight configuration, or testing the contents of `SolrDocument` or a specific Solr query.

banukutlu:
> 34332311

@ruthtillman this catkey does not exist anymore, could you find another one to check?

ruthtillman:
yeah here we go https://catalog.libraries.psu.edu/catalog/34674450 (typo resplendent)

ruthtillman:
QAd, so ready to go!

PR #826, opened 2021-08-18
Preview/search weighting

ruthtillman:
Changed minimum match threshhold and weightings on _keyword_ "pf" and "qf" fields and added title_unstem_search (245a) to keyword weighting to handle the issue of needing to weight 245a above 245ab matches.


banukutlu:
@nicktiberi would you mind for a quick review for this one, I updated some of the vcr cassettes to get the failing tests to pass due to solr changes from Ruth.

banukutlu:
> @nicktiberi would you mind for a quick review on this one, I updated some of the vcr cassettes to get the failing tests to pass due to solr changes from Ruth.

nvm CI still failing,  interestingly it is passing on my local.

solr/conf/solrconfig.xml
```diff
@@ -85,7 +85,7 @@
        <str name="echoParams">explicit</str>
        <int name="rows">10</int>
        <str name="q.alt">*:*</str>
-       <str name="mm">2&lt;-1 5&lt;-2 6&lt;90%</str>
+       <str name="mm">5lt;-1 9&lt;90%</str>
```
nicktiberi, line 88: @banukutlu - I looked into the failing tests and it seems like this change is causing the errors.

If you run the following commands on local, you will see the error if you do a search with more than 1 word in the app (e.g. `test 2` causes the error, but `test` does not).

Commands:
```
bundle exec rake solr:create_collection
bundle exec rake solr:update_config
bundle exec rake solr:create_alias
```

I don't know enough about Solr to understand what this change does, but reverting this change and running the commands again gets rid of the error.

nicktiberi: Update: I think I see the issue. An ampersand is missing from this commit. I'm gonna change it to `<str name="mm">5&lt;-1 9&lt;90%</str>` and I think that will fix it. Stay tuned...

nicktiberi (approved):
Seems good to me now!

solr/conf/solrconfig.xml
banukutlu: yes this makes sense! great catch!

banukutlu (approved):
lgtm ✔️  thanks @nicktiberi for catching the bug.

ruthtillman:
Per @banukutlu's request, I've created https://github.com/psu-libraries/psulib_blacklight/wiki/Search-Match-and-Weight-Testing and opened issue https://github.com/psu-libraries/psulib_blacklight/issues/844 for us to keep track of it, though it's not a highest-priority. 

We should discuss in next sprint when this will deploy.

PR #827, opened 2021-08-19
Availability: Remove scan request locations

### LINKED ISSUE #824: Remove scan requests for engineering, EMS, PAMS
We have been offering reserve scan requests for ENG, EMS, and PAMS library -- the libraries are set (link below). Until/unless we re-close, they're stopping offering these services. So we need to back this out but we **might** need to reinstate it if we close again. So the work should be the minimum amount to back it out while leaving the function available so we could turn it back on without too much effort. :-/ 

https://github.com/psu-libraries/psulib_blacklight/blob/63d0504d3469668d0005be6100891a4422082f00/app/javascript/availability/index.js#L525-L527

nicktiberi:
Closes #824.

Moves scan location array into `libraries_locations.json` and removes all locations from that array, effectively disabling the scan request functionality.

Adding locations back to the array will re-enable the feature for the locations in the array.

### DIFF
--- a/app/javascript/availability/index.js
+++ b/app/javascript/availability/index.js
@@ -15,6 +15,7 @@ const availability = {
     allLocations: locations.locations,
     allLibraries: locations.libraries,
     illiadLocations: locations.request_via_ill,
+    reservesScanLocations: locations.reserves_scan,
     allItemTypes: item_types.item_types,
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
this looks good to me, I have some ideas/questions re vcr for to handle related pending specs, but no need to hold up the PR, it should move forward since it is a requested change.

PR #830, opened 2021-08-20
LC Call Number Browse

awead:
Implements a call number browse feature allowing users to browse forward, backward, and search nearby call numbers.


banukutlu (approved):
lgtm 😎

PR #838, opened 2021-08-25
Use TermsQuery to parse response from Solr

awead:
Depending on the number of shards in the Solr configuration, a term query might return an array of tuples or a hash for the terms. We use the TermsQuery service to handle those differences and return a consistent array of ordered terms.


banukutlu (approved):
looks great! 👌🏻

PR #845, opened 2021-08-30
Move CatalogCleaner to PsulibBlacklight::DataManager

awead:
This fixes a bug when running rake tasks. Because RSpec is only loaded in test, the cleaner could not run because it was invoked in the dev environment.

The Cleaner class is now defined under lib, where it can be loaded either in dev or test, and separates it from catalog_cleaner.rb, where it's used only in the context of RSpec.


banukutlu (approved):
👍🏻

nicktiberi:
@awead - thanks for fixing this! I was just running a hacky fix on my local.

awead:
@banukutlu Sorry, this blew up a bit. It _should_ be fixed now. If this works, I'll update the README to reflect the new changes. @nicktiberi check this out and make sure I didn't break anything.

nicktiberi:
@awead - this works fine using the Docker setup. Thanks!

This does break launching the catalog via foreman: `bundle exec foreman start -f Procfile.dev`

But that's ok, since I'm probably the only person that still uses that and I'm fine just switching to the Docker approach for consistency. We can probably remove any remnants of the foreman stuff referenced here: https://github.com/psu-libraries/psulib_blacklight#development-setup in some other commit.

awead:
@nicktiberi We don't need to use foreman directly anymore. `bundle exec rails s` build and start webpack automatically.

lib/psulib_blacklight/cleaner.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module PsulibBlacklight
+  class Cleaner
```
banukutlu, line 4: @awead Would it be more clear if we rename this class now that it also have the loading fixtures method? I do not have suggestion though, you are way more creative for coming up with nicer names :)

awead: @banukutlu DataManager? We have a SolrManager....

banukutlu: yeas 👍🏻

awead: All fixed up!

nicktiberi:
@awead  - looks great - thank you!

One final nitpick — should we consider updating the Development Setup portion of the README now that foreman/Procfile stuff is gone?

https://github.com/psu-libraries/psulib_blacklight/blob/main/README.md#development-setup

Particularly,

Screen Shot 2021-08-31 at 11 32 08 AM

awead:
@nicktiberi Yes. I was going to do that separately. Although, I'm not 100% sure what we want to prescribe. I think we ought to move it to a wiki page where it can be edited more collaboratively. The README can have a link to the page.

nicktiberi:
> @nicktiberi Yes. I was going to do that separately. Although, I'm not 100% sure what we want to prescribe. I think we ought to move it to a wiki page where it can be edited more collaboratively. The README can have a link to the page.

Great idea! Moving dev instructions to the wiki sounds good to me. 👍

banukutlu:
yeah, we can move the dev setup section to a wiki page and link to it 👍🏻 

I was gonna mention the below needs to be updated, but this also will be part of that page then.

```
Run Tests

docker-compose exec -e RAILS_ENV=test web  bundle exec rake ci
```

banukutlu (approved):
lgtm ✔️  thanks!

as mentioned, we just need to update our dev setup instructions for readme.

PR #846, opened 2021-08-30
Availability: Add tests for <IllLink /> component

### LINKED ISSUE #832: Improve Reserves Scan Tests
Since #827, we realized that we are not testing the reserves scan scenario in the availability specs. The spec is unimplemented: https://github.com/psu-libraries/psulib_blacklight/blob/main/spec/features/availability_spec.rb#L444

We should at the very least implement a spec that hits `isReserves()`: https://github.com/psu-libraries/psulib_blacklight/blob/main/app/javascript/availability/index.js#L526-L528

Other things to consider:
- figure out a way to mock `libraries_locations.json` for the spec
- write a component spec for `location_info.jsx` to test this feature at the component level

nicktiberi:
Closes #832.

Test the various functionality of ILL link generation at the React component level, including the previously untested reserves scan functionality.

`<IllLink />` is now at 100% coverage (and somehow, so is `<Spinner />`, even though there are no tests for it...  📈 

Screen Shot 2021-08-30 at 2 16 58 PM

It might be nice to get Jest coverage in CI somehow if it isn't there already, just to make sure we don't lower our coverage going forward.

### DIFF
--- a/spec/javascript/availability/components/ill_link.jsx.spec.js
+++ b/spec/javascript/availability/components/ill_link.jsx.spec.js
@@ -1,0 +1,182 @@
+import {render, waitFor} from '@testing-library/react';
+import '@testing-library/jest-dom'
+import IllLink from '../../../../app/javascript/availability/components/ill_link';
+import availability from '../../../../app/javascript/availability';
+
+const baseUrl = 'https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/' +
+                'OpenURL?Action=10';
+
+const testLink = async (getByRole, container, label, href) => {
+  expect(getByRole('link')).toHaveAttribute('href', '#');
+  expect(container.getElementsByClassName('spinner-border').length).toBe(1);
+
+  await waitFor(() => expect(container.getElementsByClassName('spinner-border').length).toBe(0));
+
+  expect(getByRole('link')).toHaveTextContent(label);
+  expect(getByRole('link')).toHaveAttribute('href', href);
+};
+
+describe('when all fields are present', () => {
+  const moreParams = '&title=book%20title&callno=123&rfr_id=info%3Asid%2Fcatalog.libraries.psu.edu' +
+                     '&aulast=author%20name&date=2021';
+
+  beforeEach(() => {
+    $.get = jest.fn().mockImplementation(() => {
+      return Promise.resolve({
+        title_245ab_tsim: 'book title',
+        author_tsim: 'author name',
+        pub_date_illiad_ssm: 2021,
+        isbn_ssm: '1234'
+      });
+    });
+  });
+
+  test('renders an ILL link', async () => {
+    const {getByRole, container} = render(
+      <IllLink holding={{locationID: 'BINDERY', callNumber: '123'}} />
+    );
+
+    const href = baseUrl + '&Form=30&isbn=1234' + moreParams;
+
+    await testLink(
+      getByRole,
+      container,
+      'Copy unavailable, request via Interlibrary Loan',
+      href
+    );
+  });
+  
+  test('renders a reserves scan link', async () => {
+    availability.reservesScanLocations = ['BINDERY'];
+  
+    const {getByRole, container} = render(
+      <IllLink holding={{locationID: 'BINDERY', callNumber: '123'}} />
+    );
+
+    const href = baseUrl + 
+      '&Form=30&isbn=1234' +
+      '&Genre=GenericRequestReserves&location=BINDERY' + // reserves scan params
+      moreParams;
+
+    await testLink(
+      getByRole,
+      container,
+      'Copy unavailable, request via Interlibrary Loan',
+      href
+    );
+  
+    availability.reservesScanLocations = [];
+  });
+  
+  test('renders a microform link', async () => {
+    const {getByRole, container} = render(
+      <IllLink holding={{locationID: 'BINDERY', callNumber: '123', libraryID: 'UP-MICRO', itemTypeID: 'MICROFORM'}} />
+    );
+
+    const href = baseUrl +
+      '&Form=30&isbn=1234' +
+      '&Genre=GenericRequestMicroScan&location=BINDERY' + // microform params
+      moreParams;
+
+    await testLink(
+      getByRole,
+      container,
+      'Request Scan - Penn State Users',
+      href
+    );
+  });
+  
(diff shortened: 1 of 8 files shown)


spec/javascript/availability/components/ill_link.jsx.spec.js
```diff
@@ -0,0 +1,194 @@
...
+
+  expect(getByRole('link')).toHaveTextContent('Copy unavailable, request via Interlibrary Loan');
+
+  expect(getByRole('link')).toHaveAttribute('href',
+    'https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/OpenURL?' +
+    'Action=10&Form=30&isbn=&title=book%20title&callno=123' +
+    '&rfr_id=info%3Asid%2Fcatalog.libraries.psu.edu&aulast=author%20name&date=2021');
+});
+
+test('renders a link when the response comes back with no data', async () => {
```
nicktiberi, line 176: @banukutlu - what should actually happen here? Right now, if the AJAX call responds with no data, it generates and shows this seemingly incomplete link:

`https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/OpenURL?Action=10`

I'd think it might be better to just hide/remove the link altogether if there is no data. But maybe this should not be changed in this PR, because the test is at least testing what is actually happening right now.

spec/javascript/availability/components/ill_link.jsx.spec.js
```diff
@@ -0,0 +1,194 @@
...
+  await waitFor(() => expect(container.getElementsByClassName('spinner-border').length).toBe(0));
+
+  expect(getByRole('link')).toHaveTextContent('Copy unavailable, request via Interlibrary Loan');
+
+  expect(getByRole('link')).toHaveAttribute('href',
+    'https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/OpenURL?' +
+    'Action=10&Form=30&isbn=1234&title=book%20title&callno=123' +
+    '&rfr_id=info%3Asid%2Fcatalog.libraries.psu.edu&date=2021');
+});
+
```
awead, line 123: It seems like at this point you're shifting into a new "context", to borrow a term from rspec. The `beforeEach()` you have at the top is setting up a situation where you have all fields present, then from here down, you start testing different situations where there are some fields missing. However, it still looks like your original beforeEach() is being executed.

Is there a way to segment these tests so that the ones where we want all fields are separate from the others? If I was doing this in RSpec, I'd structure it like:

``` ruby
context 'when all filelds are present' do
  before { ... }

  it ...
  it ...
end

context 'when author is missing' do
  before { "do the author missing setup" }

  it ...
end
```

But, I don't know if such a setup is possible with jest.

spec/javascript/availability/components/ill_link.jsx.spec.js
```diff
@@ -0,0 +1,194 @@
...
+  expect(container.getElementsByClassName('spinner-border').length).toBe(1);
+
+  await waitFor(() => expect(container.getElementsByClassName('spinner-border').length).toBe(0));
+
+  expect(getByRole('link')).toHaveTextContent('Copy unavailable, request via Interlibrary Loan');
+
+  expect(getByRole('link')).toHaveAttribute('href',
+    'https://psu-illiad-oclc-org.ezaccess.libraries.psu.edu/illiad/upm/illiad.dll/OpenURL?' +
+    'Action=10&Form=30&isbn=1234&title=book%20title&callno=123' +
+    '&rfr_id=info%3Asid%2Fcatalog.libraries.psu.edu&aulast=author%20name');
```
awead, line 147: The link test looks really repetitious. Would it be easier to have a helper function that you could pass the parameters to and its sends back the string? You can then bake in the "https://" stuff. There might be some handy JS url builders that would make easier on the eyes.... just a thought though.

awead (commented):
Great work! I have a just a few suggestions for clarity, but otherwise this is excellent. We can look at the code climate integration later. We might need to set it up to do mulit-part submissions, which is tricky, but not too hard.

spec/javascript/availability/components/ill_link.jsx.spec.js
nicktiberi: That's a great point. Jest has `describe()` which is basically the same thing as `context` in RSpec. I'll make some changes to scope the tests as you've suggested.

spec/javascript/availability/components/ill_link.jsx.spec.js
nicktiberi: I agree. There is a lot of repetition here since all of the tests are really similar. I'll clean this up so there isn't so much code duplication.

spec/javascript/availability/components/ill_link.jsx.spec.js
banukutlu: hmm yeah that seems like a bug, we should add an issue for this.

nicktiberi: Opened a ticket for this: #848. Will remove the implementation for this test and mark it as pending/todo.

nicktiberi:
@banukutlu and @awead - this is good to go for further review.

awead:
LGTM!

spec/javascript/availability/components/ill_link.jsx.spec.js
```diff
@@ -0,0 +1,182 @@
...
+      moreParams;
+
+    await testLink(
+      getByRole,
+      container,
+      'Copy unavailable, request via Interlibrary Loan',
+      href
+    );
+  
+    availability.reservesScanLocations = [];
```
banukutlu, line 68: @nicktiberi is this for to clean this var for the rest of tests?

nicktiberi: Yep exactly. Though maybe I should store the previous value of `reservesScanLocations` and reset it, rather than setting it to an empty array at the end. This way if we add values back to `reservesScanLocations` it won't break stuff in the tests.

banukutlu (approved):
looks great! ✔️  🎉

spec/javascript/availability/components/ill_link.jsx.spec.js
banukutlu: it makes sense, just wanted to make sure if I am understanding correctly :) thanks!

PR #851, opened 2021-08-31
Use descendent_path for subject_browse_facet

awead:
Uses the same field type as `subject_facet` so they're both indexed the same way.


banukutlu (approved):
👍🏻

PR #852, opened 2021-08-31
User Login

whereismyjetpack:
- adds a login route. this path is protected in kubernetes by oauth-proxy. the proxy sends user, and header information to the application. `Settings.user_header` and Settings.groups_header` 
- Sets users groups in the session
- Only allows catalog admins to use mini profiler in production. 

To use this locally, one needs to set their request header to Settings.user_header sans `HTTP_` for example, `X_AUTH_REQUEST_EMAIL`

Visiting '/login' will give you a non-anonymous user session that you can use in the application. this will set us up for wrapping an `:admin` namespace in this constraint. The motivation for this PR was mainly to allow MiniProfiler to work in Prod when we, as admins, need to.

I still need to fiddle around with rspec to get it to test some things, but wanted to put this out for review. Testing this is hard for me to wrap my mind around for some reason.. 
This will set us up to h


config/initializers/catalog_admin.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class CatalogAdmin
```
awead, line 3: Where is this being used? I don't see it pop up anywhere.

lib/devise/models/http_header_authenticatable.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'devise/strategies/http_header_authenticatable'
+
+module Devise
+  module Models
+    module HttpHeaderAuthenticatable
+      extend ActiveSupport::Concern
+
+      def after_database_authentication; end
```
awead, line 10: I'm assuming we're noop-ing this so that Devise doesn't do its usual thing. Could you add a comment in here to confirm that?

lib/devise/redirect_to_login_failure.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class RedirectToLoginFailure < Devise::FailureApp
+  def redirect_url
+    session[:redirect_to] = request.env['ORIGINAL_FULLPATH'] unless session[:redirect_to]
```
awead, line 5: I think you want `session[:redirect_to] ||= request.env['ORIGINAL_FULLPATH']` Don't set this hash key it's already there?

lib/devise/strategies/http_header_authenticatable.rb
```diff
@@ -0,0 +1,37 @@
...
+        return pass if request.session[:user_name].present?
+
+        http_user = remote_user(request.headers)
+        return fail! if http_user.blank?
+
+        user = User.find_or_create_by(email: http_user)
+        success!(user)
+      end
+
+      def valid?
```
awead, line 17: You should be able to make this method a one liner: `remote_user(request.headers).present?`

awead (changes requested):
Looks good. We'll need to get the test coverage up, but I think this is a sound approach.

config/initializers/catalog_admin.rb
whereismyjetpack: This isn't being used yet. I stole this from scholarsphere so we can wrap routes / namespaces in a constraint. I wanted to do something like `return unless CatalogAdmin` in the minitest authorizer, but couldn't get it to work.

lib/devise/strategies/http_header_authenticatable.rb
whereismyjetpack: Good catch.

app/controllers/application_controller.rb
```diff
@@ -14,17 +14,21 @@ class ApplicationController < ActionController::Base
 
   helper_method :blackcat_config
 
+  def login
+    fullpath = request.env.fetch('ORIGINAL_FULLPATH', '/')
+    session[:redirect_url] ||= fullpath == '/login' ? '/' : fullpath
+    session[:groups] = request.env.fetch(Settings.groups_header, '').split(',')
+    redirect_to current_user ? session[:redirect_url] : '/login'
```
awead, line 21: There's a lot going on in here that ought to be unpacked a bit. First off, I'd make a private method for the redirect url. And then avoid the other ternary:

``` ruby
def login
  session[:redirect_url] ||= home_or_original_path
  session[:groups] = request.env.fetch(Settings.groups_header, '').split(',')

  if current_user?
    redirect_to session[:redirect_url]
  else
    redirect_to login_path
  end
end
```

How does it not endless redirect back to itself if current_user is not present?

app/controllers/application_controller.rb
```diff
@@ -14,17 +14,21 @@ class ApplicationController < ActionController::Base
...
   end
 
   private
 
-    # @note The Matomo ID is used to capture statistics from our production instance. This is very unlikely to change.
-    # We can use this method to limit profiling to only dev and preview instances until we can implement a way to limit
-    # it to particular users and UMGs.
     def authorize_profiler
-      return if Settings.matomo_id.to_i == 7
+      return unless request.session.fetch(:groups, []).include?(Settings.admin_group)
```
awead, line 31: Might be nice to put this is a public `admin?` method that the other controllers can access.

awead (changes requested):

app/controllers/application_controller.rb
whereismyjetpack: We could probably use that with the contraint method, then too. Where do you suppose it should live?

app/controllers/application_controller.rb
whereismyjetpack: I originally wrote with the if else, and anticipated it wasn't clever enough. :oof: 

If we run into a scenerio where there never becomes a current_user, the browser gets into a redirect loop. we could perhaps put a counter and after a threshold, redirect to a /401 -- that might be nicer to the user.

awead: There should always be a current user. It'll be a "guest user" from devise-guests.

app/controllers/application_controller.rb
awead: We're not doing route-based constraints yet, so we can just hold off and see how it shakes out.

config/initializers/catalog_admin.rb
awead: @whereismyjetpack You keep marking this as resolved, but I'm not seeing CatalogAdmin used anywhere. I think you want this in routes.rb to enable route-based restrictions, but that doesn't seem to be happening yet.

awead (changes requested):

whereismyjetpack:
This is an absolute mess. I'm going to shelf it for now. I don't have the cycles to complete it

awead:
I can fix up the test bits... I think it's pretty close.

awead:
I wasn't sure what it's supposed to do locally, anything? or does it only work when deployed?

whereismyjetpack:
If the header is set it should work, I'd be curious if we could just set it in development.rb to avoid needing to do that.

PR #854, opened 2021-09-01
Author Browse

### LINKED ISSUE #810: Author Browse Display

nicktiberi:
Closes #810.

This builds off of the existing subject browse functionality.


nicktiberi:
Have to do a bit of cleanups on this now that the prefix form is integrated. Everything is working, but I want to clean it up a bit more in the vein of @awead 's previous cleanups.

app/controllers/browse_controller.rb
```diff
@@ -17,11 +21,27 @@ def normalize_parameters
       params[:prefix].try(:capitalize!)
     end
 
+    def author_params
+      params.permit(:length, :page, :prefix)
+    end
```
awead, line 26: This can be removed.

app/components/browse/prefix_selector.rb
```diff
@@ -2,24 +2,48 @@
...
+        @action ||= params.fetch(:action) { raise KeyError, 'params must include :action' }
+      end
+
+      def authors_form_url
+        browse_authors_url
+      end
+
+      def authors_path(params)
+        browse_authors_path(params)
+      end
```
awead, line 37: remove.

app/components/browse/prefix_selector.rb
```diff
@@ -2,24 +2,48 @@
...
+
+      def subjects_form_url
+        browse_subjects_url
+      end
+
+      def subjects_path(params)
+        browse_subjects_path(params)
+      end
```
awead, line 45: remove.

app/components/browse/prefix_selector.rb
```diff
@@ -2,24 +2,48 @@
...
 
     private
 
+      def action
+        @action ||= params.fetch(:action) { raise KeyError, 'params must include :action' }
+      end
+
+      def authors_form_url
+        browse_authors_url
+      end
```
awead, line 33: I think this could be moved into the Paths module, and rename `browse_url`

app/components/browse/prefix_selector.rb
```diff
@@ -2,24 +2,48 @@
...
+
+      def authors_path(params)
+        browse_authors_path(params)
+      end
+
+      def subjects_form_url
+        browse_subjects_url
+      end
```
awead, line 41: same here: move to Paths and combine into `browse_url`

app/components/browse/prefix_selector.rb
```diff
@@ -2,24 +2,48 @@
...
     def prefix_path(letter)
-      browse_subjects_path(merge_params(prefix: letter))
+      browse_path(merge_params(prefix: letter))
     end
 
     def form_url
-      browse_subjects_url
+      raise ArgumentError, "no form URL defined for '#{action}'" unless respond_to?("#{action}_form_url", true)
+
+      send("#{action}_form_url")
```
awead, line 22: If we're doing this right, this can be removed and replaced with Paths#browse_url

app/components/browse/prefix_selector.rb
```diff
@@ -2,24 +2,48 @@
...
     end
 
     private
 
+      def action
+        @action ||= params.fetch(:action) { raise KeyError, 'params must include :action' }
+      end
```
awead, line 29: remove.

app/components/browse/prefix_selector.html.erb
```diff
@@ -19,7 +19,9 @@
     <%= form_tag form_url, method: :get, id: 'browse-form', role: 'browse' do %>
       <label for="prefix" class="sr-only"><%= t('.label') %></label>
       <div class="input-group">
-        <span class="input-group-prepend input-group-text"><%= t('.label') %></span>
+        <span class="input-group-prepend ">
```
awead, line 22: 👾

awead (changes requested):
Yeah, overall I think it's coming together. The subject search form I threw in there has added a couple of bumps that need ironing. I've noted them, but you probably already knew about them.

app/controllers/catalog_controller.rb
```diff
@@ -409,11 +409,16 @@ def show
         field.include_in_advanced_search = false
         field.label = 'Browse by Call Number'
       end
-    end
 
-    config.add_search_field('browse_subjects') do |field|
-      field.include_in_advanced_search = false
-      field.label = 'Browse by Subject'
+      config.add_search_field('browse_authors') do |field|
```
nicktiberi, line 413: @awead + @banukutlu  - I moved all of the dropdown stuff inside `unless Settings.matomo_id.to_i == 7` so none of the browse dropdown options will display in prod. Let me know if you want me to revert this.

nicktiberi:
@awead  + @banukutlu - ready for review again. 😎

app/components/browse/paths.rb
```diff
@@ -9,6 +9,10 @@ def browse_path(args)
     end
   end
 
+  def browse_url
+    send("browse_#{action}_url")
+  end
```
awead, line 14: There's a hidden bug here, but we might not run into it. `browse_url` and `browse_path` are actually methods that are defined via `ActionView::Helpers::UrlHelper` where a route defined in `routes.rb` gets the appropriate path and url methods. Currently:

```
 get '/browse', controller: :browse, action: :index 
```

Gets translated to `browse_path` and `browse_url`. What we're doing in this path module is overriding that. What would happen in the above example is that, if we were using `/browse` this would call `browse_index_url` which doesn't exist and would raise NoMethodError. However, I don't think we're using browse_url anywhere in the way it was originally intended. Still, we might want to guard ourselves against this.

Another option is to create a different method other than `browse_path` or `browse_url` and re-route accordingly.

app/components/browse/prefix_selector.rb
```diff
@@ -17,33 +17,11 @@ def prefix_path(letter)
     end
 
     def form_url
-      raise ArgumentError, "no form URL defined for '#{action}'" unless respond_to?("#{action}_form_url", true)
-
-      send("#{action}_form_url")
+      browse_url
```
awead, line 20: Can we just use `browse_url` in `form_tag` and remove this method?

awead (changes requested):

app/components/browse/paths.rb
nicktiberi: Ahh gotcha. That makes sense. Would it be sufficient to just check for the index action like you did for `browse_path` above?

```
def browse_url
  if action == 'index'
    super(args)
  else  
    send("browse_#{action}_url")
  end
end
```

awead: yeah, I think so. That's what we're doing with `browse_path`. There's probably a more elegant way once we get a named path for call numbers.

nicktiberi:
Good to go again! 🤓

awead (approved):

PR #855, opened 2021-09-01
Forms for subject browsing

awead:
This adds a drop-down option for browsing subject from a specific prefix and capitalizes any entered values so they'll work correctly with our already-capitalized facets.

It resurrects a blocked commit for call number browse because we want to do the same thing with subjects. However, call number browse needs to be hidden from non-prod hosts, so the option is blocked if we're using the Matomo ID for production.

Using the same technique we use for call numbers, we add an option for subject browsing, which redirects to the subject browse form. The PrefixSelector component is updated to include an input for supplying a custom prefix other than the canned letters.

### DIFF
--- a/app/components/browse/prefix_selector.html.erb
+++ b/app/components/browse/prefix_selector.html.erb
@@ -1,12 +1,31 @@
-<nav class="alpha-filter">
-  <ol class="pagination pagination-sm justify-content-center">
-    <li class="page-item <%= 'active' if prefix.blank? %>">
-      <%= link_to 'All', clear_prefix_path, class: 'page-link' %>
-    </li>
-    <% ('A'..'Z').to_a.each do |letter| %>
-      <li class="page-item <%= 'active' if prefix == letter %>">
-        <%= link_to letter, prefix_path(letter), class: 'page-link' %>
-      </li>
+<div class="row my-3">
+
+  <div class="col-9">
+    <nav class="alpha-filter">
+      <ol class="pagination pagination-sm mb-0">
+        <li class="page-item <%= 'active' if prefix.blank? %>">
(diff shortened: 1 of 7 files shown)


awead:
Here's what is looks like. I'm 100% sold on the UI, but this is the general idea:

Screen Shot 2021-09-01 at 2 58 45 PM
Screen Shot 2021-09-01 at 2 58 30 PM

app/components/browse/prefix_selector.html.erb
```diff
@@ -1,12 +1,31 @@
...
+        <% end %>
+      </ol>
+    </nav>
+  </div>
+
+  <div class="col-3">
+    <%= form_tag form_url, method: :get, id: 'browse-form', role: 'browse' do %>
+      <label for="prefix" class="sr-only"><%= t('.label') %></label>
+      <div class="input-group">
+        <span class="input-group-text"><%= t('.label') %></span>
```
nicktiberi, line 22: Wrap this in `<span class="input-group-prepend">` for a little nicer rendering:

### Before
Screen Shot 2021-09-01 at 4 39 05 PM

### After
Screen Shot 2021-09-01 at 4 38 48 PM

nicktiberi (changes requested):
Looking good! I am going to think about this a bit more to see if things can be moved around a bit for a cleaner UI, but I tested it and this is working well!

app/components/browse/prefix_selector.html.erb
awead: Thanks, I've added that in. I'm not sold on any of the UI stuff yet, this is just to get something up there to play with.

nicktiberi (approved):

banukutlu (approved):
LGTM 👍🏻

PR #859, opened 2021-09-02
Availability: Handle missing data for ILL + Aeon links

### LINKED ISSUE #848: Show basic ILL/Aeon links + different message if we can't retrieve data
If we can't get the data for an ILL or Aeon link, instead of showing a link with a broken URL, hide the link and maybe show an error message instead.

Relevant code: 

https://github.com/psu-libraries/psulib_blacklight/blob/main/app/javascript/availability/components/ill_link.jsx#L22-L23

If `Object.keys(data).length === 0`, then we would be showing a broken link.

nicktiberi:
Closes #848.

If we can't get JSON data for the record, we now show a basic link and a slightly different message.

We also added some more tests, including tests for `<AeonLink />`.

### DIFF
--- a/app/javascript/availability/components/aeon_link.jsx
+++ b/app/javascript/availability/components/aeon_link.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import { useEffect, useState } from 'react';
 import availability from '../index.js';
-import Spinner from './spinner.jsx';
+import SpinnerLink from './spinner_link.jsx';
 
 const AeonLink = ({holding, locationText}) => {
     const [hasData, setHasData] = useState(false);
@@ -12,6 +12,10 @@ const AeonLink = ({holding, locationText}) => {
         createAeonUrl();
     }, []);
 
+    const fetchJson = (url) => {
+        return fetch(url).then((response) => response.json());
+    };
+
     const createAeonUrl = () => {
         const catkey = holding.catkey;
         const callNumber = encodeURIComponent(holding.callNumber);
@@ -21,7 +25,7 @@ const AeonLink = ({holding, locationText}) => {
         const genre = itemTypeID === "ARCHIVES" ? "ARCHIVES" : "BOOK";
         let aeonUrl = "https://aeon.libraries.psu.edu/RemoteAuth/aeon.dll";
 
-        $.get(`/catalog/${catkey}/raw.json`).then((data) => {
+        fetchJson(`/catalog/${catkey}/raw.json`).then((data) => {
             if (Object.keys(data).length > 0) {
                 setHasData(true);
 
@@ -43,10 +47,7 @@ const AeonLink = ({holding, locationText}) => {
                     `${publisher}&ItemPlace=${pubPlace}&ItemDate=${pubDate}&ItemInfo1=${restrictions}` +
                     `&SubLocation=${subLocation}`;
             }
-
-            setShowSpinner(false);
-            setUrl(aeonUrl);
-        }).catch(() => {
+        }).catch(() => {}).finally(() => {
             setShowSpinner(false);
             setUrl(aeonUrl);
         });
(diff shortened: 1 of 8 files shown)


app/javascript/availability/components/aeon_link.jsx
```diff
@@ -18,12 +19,16 @@ const AeonLink = ({holding, locationText}) => {
         const itemID = encodeURIComponent(holding.itemID);
         const itemTypeID = holding.itemTypeID;
         const genre = itemTypeID === "ARCHIVES" ? "ARCHIVES" : "BOOK";
-        let aeonUrl = `https://aeon.libraries.psu.edu/Logon/?Action=10&Form=30` +
-            `&ReferenceNumber=${catkey}&Genre=${genre}&Location=${itemLocation}` +
-            `&ItemNumber=${itemID}&CallNumber=${callNumber}`;
+        let aeonUrl = "https://aeon.libraries.psu.edu/RemoteAuth/aeon.dll";
 
         $.get(`/catalog/${catkey}/raw.json`).then((data) => {
```
awead, line 24: What happens if this blows up with a 4/5xx?

awead (commented):
Looking good. I had a general question about error recovery. We're query our own catalog for json data, but we don't seem to be allowing for failures during that process.

app/javascript/availability/components/aeon_link.jsx
nicktiberi: @awead - great point. I can move the code to set the url + clear the spinner into a finally block to make sure we do that stuff regardless of whether the AJAX call succeeds or fails.

```
$.get().then((data) => {
  // do stuff
}).finally(() => {
  setShowSpinner(false);
  setUrl(illUrl);
});
```

nicktiberi: @awead - for some reason putting this stuff in a `finally()` block didn't behave the way I expected it to, so I added the cleanup in a `catch()` block instead. I was getting some uncaught promise rejections in Jest and it was blowing up the RSpec availability tests. Should be good for re-review.

awead: Could we just use a plain 'ol javascript Promise and not bother with jQuery?

nicktiberi: Yeah, honestly I can't see a reason why we couldn't just use `fetch()`.

fetch

awead: How fetching!

banukutlu (approved):
LGTM 👍🏻

app/javascript/availability/components/ill_link.jsx
```diff
@@ -60,15 +67,23 @@ const IllLink = ({holding}) => {
...
         return availability.illiadLocations[holding.locationID];
     };
 
+    const linkTarget = () => {
+        return hasData ? null : "_blank";
+    };
+
     return (
-        <a href={url}>
+        <a href={url} target={linkTarget()}>
```
banukutlu, line 86: @nicktiberi Wondering if  `a` href parts move to components as well?

nicktiberi: @banukutlu - we could definitely do that. Maybe this could just be called `<SpinnerLink />` or just `<Link />`? Any preference?

banukutlu: oh yeah both names makes sense actually, the component will include with the link and the spinner. I am fine with either of your suggestions, will leave it to you :)

nicktiberi:
@banukutlu  and @awead - this is ready to go again. I swapped out `$.get()` for the built-in `fetch()` and also created a `SpinnerLink` component.

banukutlu (approved):
lgtm 👌🏻  thanks @nicktiberi! great work that gets us closer to the next phase of ILL and Aeon links refactor.

PR #860, opened 2021-09-02
Refactor components

awead:
Creates a new BrowseList service to handle any facet, and also reworks our browse components a bit.


nicktiberi (approved):
Awesome cleanup - thank you!

PR #865, opened 2021-09-07
Capitalize only the first letter of the browse form prefix

### LINKED ISSUE #863: Browse Form Capitalization Issues
Any search in the browse form is capitalized, but if a user types in an all-caps search, it lowercases letters after the first, causing unexpected results:

Screen Shot 2021-09-03 at 11 42 22 AM

Screen Shot 2021-09-03 at 11 42 34 AM

We should consider attempting to only capitalize the first letter, or maybe even better, make the prefix search case-insensitive if performance wouldn't suffer.

nicktiberi:
Fixes #863.

### DIFF
--- a/app/controllers/browse_controller.rb
+++ b/app/controllers/browse_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class BrowseController < ApplicationController
-  before_action :normalize_parameters
-
   def index
     @shelf_list = ShelfListPresenter.new(shelf_list_params)
   end
@@ -16,10 +14,6 @@ class BrowseController < ApplicationController
   end
 
   private
-
-    def normalize_parameters
-      params[:prefix].try(:capitalize!)
-    end
 
     def author_list_params
       params
--- a/app/services/browse_list.rb
+++ b/app/services/browse_list.rb
@@ -14,7 +14,7 @@ class BrowseList
     @field = field
     @page = page.to_i
     @length = length.to_i
-    @prefix = prefix
+    @prefix = normalize_prefix(prefix)
   end
 
   def entries
@@ -54,5 +54,13 @@ class BrowseList
                'facet.prefix' => prefix
              })
         .dig('facet_counts', 'facet_fields', field)
+    end
+
+    def normalize_prefix(prefix)
+      return if prefix.blank?
+
+      normalized_prefix = prefix.dup
+      normalized_prefix[0] = normalized_prefix[0].upcase
+      normalized_prefix
     end
 end
--- a/app/components/browse/prefix_selector.rb
+++ b/app/components/browse/prefix_selector.rb
@@ -4,8 +4,10 @@ module Browse
   class PrefixSelector < ViewComponent::Base
     include Paths
 
-    def prefix
-      params[:prefix]
+    attr_reader :prefix
+
+    def initialize(prefix:)
+      @prefix = prefix
     end
 
     def clear_prefix_path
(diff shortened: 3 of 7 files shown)


nicktiberi:
@awead - where/how should I test this? I tried testing in `browse_controller_spec.rb`, but it didn't seem to work:

```
  describe 'capitalize_prefix' do
    it 'capitalizes the first letter of the prefix if one is present' do
      params = { prefix: 'tcp' }

      get :subjects, params: params

      binding.pry # at this point, params[:prefix] == 'tcp'
    end
  end
```

awead (commented):
I'm wondering if this should get moved into `SubjectList`. Have the list service do any necessary normalization of the prefix, and update the controller to:

``` ruby
def subjects
  list = SubjectList.new(subject_params)
  params[:prefix] = list.prefix
  @subject_list = list
end
```

Remove the `normalize_parameters` method in the controller.

nicktiberi:
@awead - I made the change and have written a spec, but am getting a `FrozenError` in the spec, which is interesting because the actual normalization is working properly. Any thoughts on why this might be happening in the spec only?

```
     Failure/Error: prefix[0] = prefix[0].upcase
     
     FrozenError:
       can't modify frozen String
```

app/controllers/browse_controller.rb
```diff
@@ -1,26 +1,24 @@
...
-  before_action :normalize_parameters
-
   def index
     @shelf_list = ShelfListPresenter.new(shelf_list_params)
   end
 
   def subjects
-    @subject_list = BrowseList.new(subject_list_params)
+    list = BrowseList.new(subject_list_params)
+    params[:prefix] = list.prefix
```
awead, line 10: On second thought, let's leave the params hash alone and make the form use `@list.prefix`. I think forcibly changing the params is kinda bad form (literally) anyway.

app/services/browse_list.rb
```diff
@@ -55,4 +55,12 @@ def facet_query
              })
         .dig('facet_counts', 'facet_fields', field)
     end
+
+    def normalize_prefix(prefix)
+      return if prefix.blank?
+
+      prefix[0] = prefix[0].upcase
```
awead, line 62: I think the params are frozen on input, so it's safer to do:

``` ruby
normalized_prefix = prefix
normalized_prefix[0].upcase
normalized_prefix
```

awead (changes requested):

app/services/browse_list.rb
```diff
@@ -55,4 +55,12 @@ def facet_query
              })
         .dig('facet_counts', 'facet_fields', field)
     end
+
+    def normalize_prefix(prefix)
+      return if prefix.blank?
+
+      normalized_prefix = prefix.dup
```
nicktiberi, line 62: @awead - I had to add `.dup` here because otherwise the string was copied by reference and not by value, so modifying `normalized_prefix` still caused the `FrozenError`. Adding `.dup` copies the string by value and avoids overwriting the original prefix string in memory.

awead: Yeah, that looks familiar.

nicktiberi:
@awead - ready for re-review.

app/services/browse_list.rb
```diff
@@ -55,4 +55,12 @@ def facet_query
              })
         .dig('facet_counts', 'facet_fields', field)
     end
+
+    def normalize_prefix(prefix)
```
banukutlu, line 59: 👌🏻 for moving this to the service from the controller.

banukutlu (approved):
LGTM 👌🏻

awead (approved):

PR #872, opened 2021-09-08
Updating our dev-related tasks

awead:
This add a new 'dev:clean' task which wipes out everything, loads any new solr configuration, and loads fixtures. It ensures that your dev environment is in a clean state and has the latest solr configuration.

Additionally, descriptions are added to the solr.rake tasks so that they'll list properly when running rake -T.

It's worth noting that we're not using the DatabaseCleaner gem to do the database truncation because it's not really worth it. Instead of introducing another dependency, we can truncate the tables via SQLite to remove the data, and rely on Rails' use_transactional_fixtures to take care the rest in between tests.

### DIFF
--- a/lib/psulib_blacklight/data_manager.rb
+++ b/lib/psulib_blacklight/data_manager.rb
@@ -17,5 +17,14 @@ module PsulibBlacklight
       Blacklight.default_index.connection.add(docs)
       Blacklight.default_index.connection.commit
     end
+
+    # @note We'd usually use a gem like DatabaseCleaner to do this, but since our database usage is so small, this
+    # seems like a simpler approach. These are SQLite commands only.
+    def self.clean_database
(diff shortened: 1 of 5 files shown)


lib/psulib_blacklight/data_manager.rb
```diff
@@ -17,5 +17,14 @@ def self.load_fixtures
       Blacklight.default_index.connection.add(docs)
       Blacklight.default_index.connection.commit
     end
+
+    # @note We'd usually use a gem like DatabaseCleaner to do this, but since our database usage is so small, this
+    # seems like a simpler approach. These are SQLite commands only.
```
banukutlu, line 22: 👍🏻

banukutlu (approved):
neat 😎

PR #873, opened 2021-09-09
load fixtures switch

whereismyjetpack:
If a deployment has the `LOAD_FIXTURES` variable set, it will run the rake task to load the fixtures. 

This change is in conjunction with a default being set on preview envs 
https://github.com/psu-libraries/catalog-config/blob/main/templates/application.yaml#L23

your specific deployment can switch this on/off as you see fit.

### DIFF
--- a/bin/startup
+++ b/bin/startup
@@ -18,6 +18,11 @@ bundle exec rails db:migrate
 echo "initalizing solr"
 bundle exec rake solr:initialize_collection
 
+if [ "${LOAD_FIXTURES:-false}" == "true" ]; then
+  echo "Loading Fixtures"
+  bundle exec rake solr:load_fixtures
+fi
+
 echo "starting rails"
 
 bundle exec rails s -b '0.0.0.0'


awead (commented):
Works for me. The only thing that's triggering my spidey-sense is that we should never do this in production-production. It wouldn't break anything, but it could cause some records to get overwritten with potentially old data. I guess it just means being careful in the config repo.

whereismyjetpack:
yeah, the default is turned off, so we'd have to be explicit in turning it on in qa/prod in the config repo, which could happen, but isn't the default.

awead:
@whereismyjetpack how is the default turned off? In https://github.com/psu-libraries/catalog-config/commit/633e8e5d1bf2045463d0e7233e3fcd9a25a8acc3 it appears to be on. Is the default "true" for preview branches and false for others? If not, do we need to update the yaml for the specific deployment?

whereismyjetpack:
That template only gets picked up by previews, the default values are false. 

https://github.com/psu-libraries/catalog-config/blob/main/charts/psulib_blacklight/values.yaml#L13

PR #875, opened 2021-09-09
Don't use descendent_path for subject browsing

awead:
Subject facets used in browsing should not be split on the em dash, which Solr does via the descendent_path field type. We've added an additional marc record the test set in order to verify this behavior and prevent regressions.

Fixes psu-libraries/psulib_traject#354


awead:
I pushed this to a preview branch, but if it looks good, we can just merge to QA and see what it does with the real data.

banukutlu (approved):
lgtm 👍🏻

PR #876, opened 2021-09-09
Add options for including lower term

### LINKED ISSUE #831: Call no. display is duplicating first result
Seeing same record twice at the top while moving via Next (and maybe on landing page?) e.g.: https://catalog-call-number-browse.dev.k8s.libraries.psu.edu/browse?starting=NK.0460.C7.B8

```
NK460.C7B8 	The family collections at Van Cortlandt Manor / [by] Joseph T. Butler 	Annex / Penn State Harrisburg 	Tarrytown, N.Y. : Sleepy Hollow Restorations, [1967]
NK460.C7B8 	The family collections at Van Cortlandt Manor / [by] Joseph T. Butler 	Annex / Penn State Harrisburg 	Tarrytown, N.Y. : Sleepy Hollow Restorations, [1967]
```

which are actually the same record, not two very similar records.

(Note: when I switch to paging backward using Previous, it's fine -- I'm not seeing duplicates on either end of the results, but then it happens again when I start using Next)

awead:
When retrieving shelf lists with TermsQuery, we only need to include the lower term, which is the query term or book we're looking for, in either the forward or reverse set, but not _both_. Including it in both results in the book present in the forward and reverse set. When the shelf is then assembled with both queries, the book will appear twice.

Adding an option to include the lower term allows us to only request it for the reverse set and not the forward set. This creates a virtual shelf with the queried item only appearing once instead of twice.

Fixes #831

### DIFF
--- a/app/services/shelf_list.rb
+++ b/app/services/shelf_list.rb
@@ -41,7 +41,8 @@ class ShelfList
       @forward_keys ||= TermsQuery.call(
         field: forward_shelfkey,
         limit: forward_limit,
-        query: query
+        query: query,
+        include_lower: 'false'
       )
     end
 
--- a/app/services/terms_query.rb
+++ b/app/services/terms_query.rb
@@ -1,16 +1,17 @@
 # frozen_string_literal: true
 
 class TermsQuery
-  def self.call(field:, limit:, query:)
-    new(field, limit, query).terms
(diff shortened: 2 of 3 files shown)


banukutlu (approved):
makes sense! thanks @awead!

PR #879, opened 2021-09-10
Add docValues to subject_browse_facet

awead:
Subject browsing still doesn't seem to behave correctly. This change makes the field configured _exactly_ like the others, so hopefully it'll work this time.

Fixes psu-libraries/psulib_traject#354

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -229,7 +229,7 @@
     <field name="library_facet"        type="string" docValues="true" indexed="true" stored="false" multiValued="true" />
     <field name="location_facet"       type="string" docValues="true" indexed="true" stored="false" multiValued="true" />
     <field name="media_type_facet"     type="string" docValues="true" indexed="true" stored="false" multiValued="true" />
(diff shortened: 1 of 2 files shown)


awead:
The problem is that this issue doesn't appear locally, only in QA. I suspect it's because QA's Solr config uses shards. But for whatever reason, the facet query for `Quilting—Pennsylvania—Cumberland County—History—19th century` under the `subject_browse_facet` field shows no results. But, it _does_ show results if I use the `subject_facet` field. The latter uses the descendent_path type, which we don't want, so that only thing left seems to be to make the subject_facet_field exactly like the other non-descendent_path fields. 🤷‍♂️

banukutlu (approved):
worth trying 🤞🏻 

maybe we should consider sharding in our local setup as well 🤔

awead:
I asked @whereismyjetpack but that's reeeeally complicated because you need multiple nodes?

banukutlu:
> I asked @whereismyjetpack but that's reeeeally complicated because you need multiple nodes?

hmm that is right

whereismyjetpack:
You should be able to set shard count higher. You need to also set a flag to allow it

PR #886, opened 2021-09-15
Remove CGI.escape() from redirect_browse

### LINKED ISSUE #885: We lost Ursula Le Guin? (spaces / separators not appearing to search correctly?)
Identified the following issue with drop-down search box with Author Browse and Subject Browse: When searching for a term, the spaces (and other separators like commas?) aren't behaving as expected. e.g. `Roosevelt` works but `Roosevelt, Anna` does not. This is also affecting compound names like `Le Guin` which means that it's not just commas but also spaces.

Searches for the following:
```
Le Guin, Ursula K., 1929- (this would be unlikely but might be pasted in)
Le Guin, Ursula
Le Guin, U*
Le Guin, U*
Le Guin
```
should all yield at least some results but are all returning "No records found." The first four should return 2 results (there's 2 forms of her name, that's a thing we can identify and fix with better browse view. The fifth should return 5.

In Subject Browse, we're getting a match for:

`Quilting--Australia--History`

but when there's a space, e.g.  `Quilting--New Mexico`  (or `Assassins--Computer games`) then it's not working. `Assassins--Computer` does work, as does `Quilting--New` so they're prefixing ok, just not matching when there's a space.

nicktiberi:
Fixes #885.

The search query was already being escaped properly, and adding `CGI.escape()` on top of that was actually causing errors.

### DIFF
--- a/app/controllers/concerns/browse.rb
+++ b/app/controllers/concerns/browse.rb
@@ -8,17 +8,13 @@ module Browse
 
     case params[:search_field]
     when 'browse_cn'
-      redirect_to browse_path(nearby: search_params)
+      redirect_to browse_path(nearby: params[:q])
     when 'browse_authors'
-      redirect_to browse_authors_path(prefix: search_params)
+      redirect_to browse_authors_path(prefix: params[:q])
     when 'browse_subjects'
-      redirect_to browse_subjects_path(prefix: search_params)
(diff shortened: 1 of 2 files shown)


awead (approved):
Nice. Yeah, Rails does that for us...

banukutlu (approved):
👍🏻

banukutlu (approved):
👍🏻

PR #890, opened 2021-09-17
Browse by Dewey Call Numbers Display

### LINKED ISSUE #864: Browse Dewey Call Numbers
Browse Dewey call numbers like we do with LC call numbers

## WDLL

* drop-down menu option for Dewey
* forward and reverse browsing
* nearby browsing
* all of the other features in LC browse
* update paths to be:
    * `/browse/lc`
    * `/browse/dewey`

banukutlu:
Closes #864


app/controllers/browse_controller.rb
```diff
@@ -2,7 +2,11 @@
 
 class BrowseController < ApplicationController
   def index
-    @shelf_list = ShelfListPresenter.new(shelf_list_params)
+    @shelf_list = ShelfListPresenter.new(lc_shelf_list_params)
+  end
+
+  def dewey
+    @shelf_list = ShelfListPresenter.new(dewey_shelf_list_params)
```
awead, line 9: Take a look at https://github.com/psu-libraries/psulib_blacklight/pull/891. I think it'll be easier if we treat classification as a parameter, ex: `/browse/call_numbers?classification=lc`. However, we might want to make it a kind of enum.

app/components/browse/call_number_navigation.rb
```diff
@@ -5,11 +5,11 @@ class CallNumberNavigation < Navigator
     delegate :next_item, :previous_item, to: :list
 
     def previous_path
-      browse_path(ending: previous_item.key, length: length)
```
awead, line 8: add a helper method

``` ruby
def classification
  params[:classification]
end
```

this becomes:

``` ruby
browse_path(ending: previous_item.key, length: length, classification: classification)

File: app/controllers/browse_controller.rb [hunk 3]
```diff
@@ -31,6 +35,14 @@ def shelf_list_params
...
     end
 
+    def lc_shelf_list_params
+      shelf_list_params.merge(classification: 'lc')
+    end
+
+    def dewey_shelf_list_params
+      shelf_list_params.merge(classification: 'dewey')
+    end
+
```
awead, line 45: add `:classification` to the list of permitted params and you can use the original method.

app/models/holdings.rb
```diff
@@ -38,8 +38,9 @@ class Holdings
   # @param documents [Array<SolrDocument, Hash>] Solr documents or hash objects from a query
   # @param direction [String] The direction in which we're looking, either 'forward' or 'reverse'
   # @note Direction is used so that we don't add every shelf key to the items hash.
-  def initialize(documents:, direction:)
+  def initialize(documents:, direction:, classification:)
     @items = {}
+    @classification = classification
```
awead, line 43: If we add classification to the url params, does Holdings need it anymore? I'm asking because I genuinely don't know :)

app/models/shelf_item.rb
```diff
@@ -5,16 +5,17 @@
 # ShelfList service.
 
 class ShelfItem
-  attr_reader :call_number, :documents, :key
+  attr_reader :call_number, :documents, :key, :classification
   attr_accessor :match, :nearby
 
-  def initialize(call_number:, key:, label: nil, nearby: false)
+  def initialize(call_number:, key:, label: nil, nearby: false, classification: 'lc')
```
awead, line 11: Same here. I think the only reason we need to know it is for the components. But if it's in the params hash... 🤷

app/services/shelf_list.rb
```diff
@@ -12,20 +12,20 @@ def self.call(args)
...
     reverse_limit:,
-    forward_shelfkey: 'forward_lc_shelfkey',
-    reverse_shelfkey: 'reverse_lc_shelfkey'
+    classification:
   )
     @query = query
     @reverse_limit = (reverse_limit || 10).to_i
     @forward_limit = (forward_limit || 10).to_i
     @forward_shelfkey = forward_shelfkey
     @reverse_shelfkey = reverse_shelfkey
```
awead, line 27: you don't need the forward_shelfkey and reverse_shelfkey attr_readers or these lines above if you have the methods below.

app/services/shelf_list.rb
```diff
@@ -35,6 +35,14 @@ def build
     }
   end
 
+  def forward_shelfkey
+    "forward_#{classification}_shelfkey"
+  end
+
+  def reverse_shelfkey
+    "reverse_#{classification}_shelfkey"
+  end
```
awead, line 44: I had originally intended for the shelfkey fields to be passed in during initialization so that this class didn't need to know about how they were named, but I completely understand why you need this now.

app/services/shelf_list.rb
```diff
@@ -50,7 +58,7 @@ def reverse_keys
       @reverse_keys ||= TermsQuery.call(
         field: reverse_shelfkey,
         limit: reverse_limit,
-        query: ShelfKey.reverse(query)
+        query: ShelfKey.reverse(query, classification)
```
awead, line 61: Technically, we don't _have_ to know the classification in order to reverse it. We can just reverse whatever the query is. We'd need to have another method for "plain reverse" or re-think the implementation.

app/views/browse/dewey.html.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Browse by Dewey Call Number</h1>
```
awead, line 1: You'd rename this partial to `_dewey` if we put classification as a parameter and not in the named path.

bin/attach
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+docker attach $(docker ps -f name=psulib_blacklight_web_1 -q)
```
awead, line 2: this might have snuck in there for some other reason

awead (changes requested):
I suggested a different tack with url parameters that I think might simplify this and remove the need to pass classification around quite so much. See the PR, then maybe give it a try and let me know how it feels.

app/components/browse/call_number_navigation.rb
banukutlu: done

app/controllers/browse_controller.rb
banukutlu: done

bin/attach
banukutlu: I added this on purpose, find it handy to attach to the running container with just `bin/attach` for when debugging with pry or so.

app/views/browse/dewey.html.erb
banukutlu: done

app/models/holdings.rb
banukutlu: nope not needed after changing classification as a url param

app/models/shelf_item.rb
banukutlu: removed

app/services/shelf_list.rb
banukutlu: removed

bin/attach
awead: I think `docker-compose web` will attach to the current web container. But I'm curious why you would need that for debugging. Are you running your tests inside the container?

banukutlu: Correct, it attaches to the current so for debugging. And yes I am running tests inside the container.

awead: Running the tests outside of docker is much faster, in my experience. The only thing I use docker for is Solr. I run redis locally, although you don't have to. We should talk maybe? I can show you how I have it set up. I wrote up an updated dev setup here: https://github.com/psu-libraries/psulib_blacklight/wiki/Development-Setup

banukutlu: right, it is a bit slower. I will try to go with the way instructions are and see, thanks 👍🏻

app/controllers/browse_controller.rb
banukutlu: I have added changes to make it a param.

bin/attach
awead: Sure thing. Please bug me if you have problems.

banukutlu: sure, thanks.

I also removed this.

app/services/shelf_list.rb
banukutlu: I got rid of classification from ShelfKey and moved it the prefix check to ShelfList, pls let me know what you think of that.

awead (dismissed):

awead:
@banukutlu Looks good. There's a merge conflict from Nick's earlier PR.

banukutlu:
> @banukutlu Looks good. There's a merge conflict from Nick's earlier PR.

rebased 👍🏻

awead (approved):

PR #896, opened 2021-09-23
hold button url default to myaccount qa

banukutlu:
we will only need to override it for myaccount prod - https://github.com/psu-libraries/catalog-config/blob/main/clusters/prod/manifests/psulib_blacklight/prod.yaml#L50


whereismyjetpack:
sorry for the back and forth, i think this makes a lot more sense, too!

whereismyjetpack (approved):

banukutlu:
> sorry for the back and forth, i think this makes a lot more sense, too!

np, and it is on me as well :) thanks for your help, this way makes more sense to me, too and less overriding :)

PR #897, opened 2021-09-23
Use shelvit gem to convert user call number input into shelf keys

### LINKED ISSUE #888: Call Number Browse: Query for nearby items by shelf key
Currently, we are querying for nearby items by their base call number, but really we should transform this call number into a normalized shelf key.

Traject does this normalization here: https://github.com/psu-libraries/psulib_traject/blob/main/lib/psulib_traject/shelf_key.rb#L23

We could do something like this in `shelf_list_presenter.rb`:

```diff
...

   def initialize(params = {})
     @starting = params[:starting]
     @ending = params[:ending]
-    @nearby = params[:nearby]
+    @nearby = Catify.normalize(params[:nearby])
     @length = params.fetch(:length, DEFAULT).to_i
   end

...

-      if before_list.last.call_number == nearby
+      if before_list.last.key == nearby

...
```

It would be nice to make this into a small gem so we can use the same normalization logic in both Traject and Blackcat.

Related comment from Ruth: https://github.com/psu-libraries/psulib_blacklight/issues/800#issuecomment-920190196

nicktiberi:
Closes #888.

This should improve the browse experience since we are now querying on normalized shelf keys instead of the base call numbers.

### DIFF
--- a/app/presenters/shelf_list_presenter.rb
+++ b/app/presenters/shelf_list_presenter.rb
@@ -56,6 +56,10 @@ class ShelfListPresenter
       (before_items + after_items)
     end
 
+    def shelf_key
+      @shelf_key ||= Shelvit.normalize(nearby) || nearby
+    end
+
     # @return [Array<ShelfItem>]
     # @note The items coming before a particular call number are determined using a reverse shelf key, which means the
     # items are returned in a "reverse" order. To create a natural reading list, we need to reverse the reversed list
@@ -65,9 +69,9 @@ class ShelfListPresenter
     def before_items
       before_list = shelf_list[:before].reverse
 
-      return before_list if nearby.blank?
+      return before_list if shelf_key.blank?
 
-      if before_list.last.call_number == nearby
+      if before_list.last.key == shelf_key
         before_list.last.match = true
       else
         before_list << ShelfItem.new(label: "You're looking for: #{nearby}", call_number: 'None', key: nil)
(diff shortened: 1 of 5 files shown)


banukutlu (dismissed):
LGTM ✅  🚀

app/presenters/shelf_list_presenter.rb
```diff
@@ -16,7 +16,8 @@ class ShelfListPresenter
   def initialize(params = {})
     @starting = params[:starting]
     @ending = params[:ending]
-    @nearby = params[:nearby]
+    @label = params[:nearby]
+    @nearby = Shelvit.normalize(params[:nearby]) || params[:nearby]
```
awead, line 20: I'd probably make this a method:

``` ruby
def nearby
  @nearby ||= Shelvit.normalize(params[:nearby]) || params[:nearby]
end
```

app/presenters/shelf_list_presenter.rb
```diff
@@ -67,10 +68,10 @@ def before_items
 
       return before_list if nearby.blank?
 
-      if before_list.last.call_number == nearby
+      if before_list.last.key == nearby
         before_list.last.match = true
       else
-        before_list << ShelfItem.new(label: "You're looking for: #{nearby}", call_number: 'None', key: nil)
+        before_list << ShelfItem.new(label: "You're looking for: #{@label}", call_number: 'None', key: nil)
```
awead, line 74: `attr_reader :label`

awead (changes requested):
Couple of small points. I wonder if we need to change up the parameters. We seem to be confusing "nearby" and "label". We might want to keep the `nearby` method always as `param[:nearby]` and the normalized parameter as a new method.

app/presenters/shelf_list_presenter.rb
awead: See my earlier comment. This might ought to be a different method. The param and the method could be returning different things which might be confusing.

nicktiberi: What if we keep `@nearby = params[:nearby]` and then make a `shelf_key` method?

```ruby
def shelf_key
  Shelvit.normalize(nearby) || nearby
end
```

awead: works for me! Memoize it, if gets called more than once.

banukutlu (approved):
approving to get this merged, so I will get these changes to the dewey branch.

thanks Nick

awead (approved):

PR #920, opened 2021-12-01
Subject and Author Browse access public on production

### LINKED ISSUE #915: Update access levels to call number browse before Sunset
We'll be removing access restrictions on browse author & subject and but keep restrictions on browse call number and everything will be added to the dropdown and unhide the nearby on shelf on item view (that is with availability section)

banukutlu:
Closes #915


nicktiberi (approved):
👍

awead (approved):

PR #923, opened 2021-12-03
Removes genre_full_facet from CatalogController

### LINKED ISSUE #893: Both Genre Facet and Genre Full Facet are appearing on search results page
edited for clarity: Can we remove the genre_full_facet box from the sidebar? example https://catalog.libraries.psu.edu/catalog?commit=Search&op=AND&all_fields=&title=moby+dick&author=&subject=&identifiers=&series=&publisher=&range%5Bpub_date_itsi%5D%5Bbegin%5D=&range%5Bpub_date_itsi%5D%5Bend%5D=&sort=score+desc%2C+pub_date_itsi+desc%2C+title_sort+asc&search_field=advanced

awead:
The field is not used for facet navigation, but is used for facet linking. So other methods than interact with it for url linking purposes need to remain, but it should appear as an actual facet on the left-side navigation pane.

Fixes #893

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -179,7 +179,6 @@ class CatalogController < ApplicationController
     config.add_facet_field 'language_facet', label: 'Language', limit: true
     config.add_facet_field 'subject_topic_facet', label: 'Subject', limit: 20, index_range: 'A'..'Z'
     config.add_facet_field 'genre_facet', label: 'Genre', limit: 20, index_range: 'A'..'Z'
-    config.add_facet_field 'genre_full_facet', label: 'Genre', limit: 20, index_range: 'A'..'Z'
     config.add_facet_field 'media_type_facet', label: 'Media Type', limit: 20, index_range: 'A'..'Z'
     config.add_facet_field 'classification_pivot_field',
                            label: 'Call Number',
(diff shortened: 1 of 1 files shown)


nicktiberi:
Removing this from the catalog config breaks genre links.

For example, on this record: https://catalog.libraries.psu.edu/catalog/22080733

Clicking one of the genre links takes us to https://catalog.libraries.psu.edu/?f[genre_full_facet][]=Domestic+fiction

If `genre_full_facet` is removed from the config, these links no longer narrow down the search results properly.

awead:
@nicktiberi Right you are. I asked about this in #893. Make sure your concept of "breaking" matches mine, but it looks like this is what we want to do.

nicktiberi:
I'm not sure if our definition of breaking is matching up. In addition to the facet selection not appearing on the left nav (which seems fine), the results will not be filtered properly if `genre_full_facet` is removed from the config.

http://localhost:3000/?f[genre_full_facet][]=Domestic+fiction

### genre_full_facet in config (results filtered by genre):
Screen Shot 2021-12-07 at 2 34 41 PM

### genre_full_facet removed from config (results not filtered by genre):
Screen Shot 2021-12-07 at 2 35 05 PM

awead:
@ruthtillman can you comment ☝️ ?

awead:
I think the difference here is that clicking on the "Genre" field isn't a faceted search. Instead it's just a field search, i.e. "Show me all the records that have `x` as their genre". There should be a way to do this just by modifying the href link in genre.

ruthtillman:
I think that's probably how we accidentally ended up with two of them in the sidebar... but whatever the method we use it will probably resemble whatever we're doing with all_authors (linking? other thing?) because that's a similar kind of facet-type search AFAIK but only present in record links but not in the sidebar.

nicktiberi:
> I think that's probably how we accidentally ended up with two of them in the sidebar... but whatever the method we use it will probably resemble whatever we're doing with all_authors (linking? other thing?) because that's a similar kind of facet-type search AFAIK but only present in record links but not in the sidebar.

Ah! If we add `show: false` in the config, this preserves the filtering when clicking the genre links, but also hides the duplicate genre facet from the sidebar:

```
config.add_facet_field 'genre_full_facet', label: 'Genre', limit: 20, index_range: 'A'..'Z', show: false
```

awead:
Yeah, that's probably what we can do for now, but we should find another way to have this behavior without using facets because the the hidden facets are still in the solr responses and that can adversely affect performance.

nicktiberi:
FWIW, this comment seems to imply that setting `show: false` makes it so that these facets are NOT in the Solr response. Not sure if that is actually true or not...

https://github.com/psu-libraries/psulib_blacklight/blob/main/app/controllers/catalog_controller.rb#L194

```
#
# Facets that are configured but are not in the solr response
#
config.add_facet_field 'all_authors_facet', show: false
```

nicktiberi (dismissed):

awead:
If I do a search like: https://catalog.test/?f%5Bformat%5D%5B%5D=Book and tail the `log/development.log`, you'll see `genre_full_facet` under `facet.field`. However, I think if we set limit to zero, there won't be any data in the facet data and it'll still make the search work.

nicktiberi (approved):
Sorry for all the back and forth on this small change. Thanks for taking care of this! 🎉

PR #925, opened 2021-12-07
Show summary holdings info in the UI

### LINKED ISSUE #419: Summary Holdings display
(Ruth edit 2021-08-31)

Functionally, summary holdings provide information "at a glance" regarding our journals, magazines, newspapers, and other periodicals. For example, if we have over 100 volumes, the summary holdings could provide you with the basic, helpful information: 1893-1922; 1931-1935; 1937-1992; Indexes to 1980-1990. --- which helps you know you won't find anything more recent than 1992 and there are some gaps. They also explain what a call number like "Periodical" means -- e.g. that we have "Current issue" or "Current year only."

Once we've [indexed them](https://github.com/psu-libraries/psulib_traject/issues/348) we need to determine how we'll display them. The current [model used in the CAT](https://github.com/psu-libraries/psulib_blacklight/wiki/Summary-Holdings-Display#how-it-looks-in-the-cat) can almost certainly be improved on!

Tentative idea: In Availability, match library codes and display the summary holdings at the top of each library's section in availability.

We could skip over re-displaying the library name and call number but would want to match the location code to our translation table so that we could differentiate in repeating cases. For simplicity's sake, we wouldn't want to display the location unless there was more than one.

Notes:

- If there are 866, 867 or 868 without a 852 to group by, we will not display those. 
- There might be 852 with only 867 or 868 as well, we will display what we can find for that 852. 
- Each 866, 867, 868 will be their each own row when displaying.

See wiki: https://github.com/psu-libraries/psulib_blacklight/wiki/Summary-Holdings-Display

nicktiberi:
Closes #419.

Screen Shot 2021-12-07 at 10 01 54 AM


awead (dismissed):

awead (approved):
Go. Go. Go. Go. Go. Go. Go. Go...

Issue #766, opened 2021-06-16
Change notes field concatenation.


ruthtillman:
Right now, repeating 5xx (note) fields are concatenated using `,` and `and`. Because notes fields are often sentences or paragraphs, this leads to a really awkward public display. We should probably change them all to either repeat or to be separated in display by a line break at least. 

Examples where it's currently weird: 

- https://catalog.libraries.psu.edu/catalog/2171372
- https://catalog.libraries.psu.edu/catalog/2018291
- https://catalog.libraries.psu.edu/catalog/6520236

(I'll add others as I see them)

ruthtillman:
@nicktiberi per our discussion, these are the ones which I've identified which would be better with line breaks... which is a lot of them -- reasoning that they're often formatted like sentences/paragraphs and thus behave weirdly when concatenated.

```
'overall_imprint_display_ssm'
'copyright_display_ssm'
'edition_display_ssm'
'phys_desc_ssm'
'series_title_display_ssm'
'language_note_ssm'
'restrictions_access_note_ssm'
'notes_summary_ssim'
'serials_formed_from_display_ssim'
'serials_absorbs_display_ssim'
'serials_absorbed_by_display_ssim'
'serials_absorbs_in_part_display_ssim'
'serials_absorbed_in_part_by_display_ssim'
'serials_separated_from_display_ssim'
'serials_split_into_display_ssim'
'serials_merged_to_form_display_ssim'
'serials_changed_back_to_display_ssim'
'dates_of_pub_ssim'
'isbn_ssm'
'issn_ssm'
'related_title_display_ssm'
'duration_ssm'
'frequency_ssm'
'collection_facet'
'finding_aid_note_ssm'
'provenance_note_ssm'
'dissertation_note_ssm'
'audience_ssm'
'reading_grade_ssm'
'interest_age_ssm'
'interest_grade_ssm'
'sound_ssm'
'music_numerical_ssm'
'music_format_ssm'
'music_key_ssm'
'performance_ssm'
'video_file_ssm'
'scale_graphic_material_note_ssm'
'cartographic_mathematical_data_ssm'
'digital_file_ssm'
'form_work_ssm'
'audience_notes_ssm'
'general_note_ssm'
'other_edition_ssm'
'bibliography_note_ssm'
'with_note_ssm'
'creation_production_credits_ssm'
'citation_references_note_ssm'
'participant_performer_ssm'
'type_report_period_note_ssm'
'special_numbering_ssm'
[...]

PR #929, opened 2021-12-10
Display each report number on its own line for readability

nicktiberi:
Re: #766.

Use the `newline_format` helper to display report numbers separated by line breaks instead of in a comma-separated list.


banukutlu (approved):
lgtm ✔️  thanks Nick!

PR #936, opened 2021-12-15
updates node in CI

whereismyjetpack:
This looks crazy train, but it's how Circle recommends us do it, and it's just CI _shrug_


nicktiberi (approved):
🥇

PR #938, opened 2021-12-15
Show dedication and endowment info for catalog records in the UI

### LINKED ISSUE #57: Display "digital bookplates" (just text for now) and dedications
Ruth updated 2021-11-16 with current practices and mapping so things below may apply less. We will currently not display actual digital bookplates, but the data we're creating through indexing will let us create image bookplates/enhanced search/however we want to do it. See the wiki https://github.com/psu-libraries/psulib_blacklight/wiki/Endowments-and-Dedications for big picture and https://github.com/psu-libraries/psulib_traject/issues/12 for indexing.

## Digital Bookplates

Digital Bookplates should be displayed as follows (pseudocode):

```
Acquired with generous support from the {link to controlled search for {link to controlled search for 993a}{993n}{/link}
```

We currently don't use a label for them and don't seem to for now. These may repeat and should be _displayed on separate lines_.

Note that the book _may or may not_ have a similar 541 field, e.g. https://catalog.libraries.psu.edu/catalog/4322824 but the number is different than in the endowments link above.

## Dedications

These may repeat and should be _displayed on separate lines_.

| label | value |
| Dedication | 995a |

nicktiberi:
Closes #57.

Screen Shot 2021-12-15 at 11 31 53 AM

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -359,6 +359,8 @@ class CatalogController < ApplicationController
     config.add_show_field 'indexed_by_note_ssm', label: 'Indexed By', helper_method: :newline_format
     config.add_show_field 'selectively_indexed_by_note_ssm', label: 'Selectively Indexed By', helper_method: :newline_format
     config.add_show_field 'references_note_ssm', label: 'Reviewed/Cited In', helper_method: :newline_format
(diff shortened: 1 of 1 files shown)


banukutlu (approved):
lgtm 👌🏻

PR #939, opened 2021-12-15
solr:delete_collection rake task

whereismyjetpack:
Presently as part of our workflow, we will delete the collection in the solr admin, or by firing off a `curl` call 

delete_collection will delete the passed collection out of solr. if the collection isn't found, all collections that are in solr are emitted to screen so we can easily copy and paste. 

If we attempt to delete a collection that has an alias, solr will return an error.

### DIFF
--- a/lib/psulib_blacklight/solr_manager.rb
+++ b/lib/psulib_blacklight/solr_manager.rb
@@ -67,6 +67,20 @@ module PsulibBlacklight
       end
     end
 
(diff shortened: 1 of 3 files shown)


banukutlu (commented):
this looks great 👍🏻  just adding a spec will do it.

thanks so much for this @whereismyjetpack, will be very useful 👌🏻

banukutlu (approved):
👍🏻

Issue #759, opened 2021-06-14
Access Online links at the bottom of the item display


banukutlu:

ruthtillman:
So the reason we put these at the bottom was because they're not as relevant as the main Access Online link _and_ there may be a ton of them so links which aren't to the actual item but to a related item _might_ stand between a user and the Availability section or the user and the catalog record data.

Display ideas:

1. Put these links directly below the Access Online links. If there are more than -- 3? 5? -- then collapse.
2. Put these links at the top of the fields below Availability. If there are more than -- 3? 5? -- then collapse.

Also will add to the indexing -- when updating all this it would also be reasonable to add notes/labels so folks understand the kind of relationship.

ruthtillman:
Change label for `partial_links_struct` to `Partial Online Access` because it could be particularly confusing if we move it farther up and we need to differentiate between it and Access Online (plus it's not the clearest right now anyway)

https://github.com/psu-libraries/psulib_blacklight/blob/769a356858d78f0dc25f40b3c23757e66c6e9813/app/controllers/catalog_controller.rb#L326

ruthtillman:
I'll get on it--it may take a little while though!

banukutlu:
@ruthtillman  links updates preview https://catalog-links.dev.k8s.libraries.psu.edu/, should have all changes except the "if more than 3 links collapse" part.

PR #941, opened 2022-01-04
External links refactoring

banukutlu:
Part of #759 

This is to move all external links to the top section to display below the Access Online links as mentioned [here](https://github.com/psu-libraries/psulib_blacklight/issues/759#issuecomment-866189860), also includes a bit of refactoring to change the external links partials to components. This is more of a first pass since these components can be more cleaned up.

There is more todo in terms of the display of the links aka collapsing if there are more than 3 or 5 links. I wanted to get some feedback before adding more changes.


awead (changes requested):
I really like where this is headed. I've added some suggestions inline in the diff, but that last thing I want to suggest is that I think there's some room for reducing the number of link component classes. I think you could get away with three or less. Define a base class:

``` ruby
module ExternalLinks
  class BaseLinksComponent < ViewComponent::Base
    attr_reader :heading

    def initialize(links:, heading:, show_links: true)
      @links = links
      @heading = heading
      @show_links = show_links
    end
    
    def render?
      links.present? && show_links?
    end

    private
      attr_reader :links
      
      def show_links?
        !!@show_links
      end
  end
end
``` 

Then the view is:

``` ruby
<div class="bs-callout bs-callout-primary">
  <h5><%= heading %></h5>

  <ul class="external-links-access-online">
    <% links.map do |item| %>
      <%= render ExternalLinks::LinkComponent.new(item) %>
    <% end %>
  </ul>
</div>
```

That should cover all of them except Hathi and PsuDigitalCollections. For those, define subclasses for your base:

``` ruby

class HathiLinkComponent < BaseLinksComponent
  def initialize(links:, show_component:)
    @links = links
    @show_component = @show_component
  end
end

class PsuDigitalCollectionsLinksComponent < BaseLinksComponent
end
```

And then have separate html view only for those those two.

It might be overkill... 🤷 But something to play around with.

Another idea is to add the image_tag as an option to the base class and expand the LinkComponent to include the additional hathi keys, then you could have only _one_ links component class for everyting, but that seems a little forced.

awead:
GREAT MINDZ!!!! 🧠 <=> 🧠

app/components/external_links/psu_digital_collections_links_component.html.erb
```diff
@@ -6,8 +6,12 @@
...
 
   <ul class="external-links-psu-digital-collections">
-    <% psu_digital_collections_links.map do |item| %>
-      <li><i class="fas fa-external-link-alt" aria-hidden="true"></i> <%= item['prefix'] %><a href="<%= item['url'] %>"><%= item['text'] %></a><%= item['notes'] %></li>
+    <% links.map do |item| %>
+        <li>
+          <i class="fas fa-external-link-alt" aria-hidden="true"></i> <%= item['prefix'] %>
+          <%= link_to item['text'], item['url'] %>
+          <%= item['notes'] %>
+        </li>
```
awead, line 14: Use the ExternalLinks::LinkComponent here.

awead (changes requested):
Fantastic! That really came together nicely. You missed one thing in the collections links component. Otherwise, this is SOLID.

app/components/external_links/psu_digital_collections_links_component.html.erb
banukutlu: ah yes, thanks!

nicktiberi:
This looks great! Love seeing our usage of components go up. 🚀 

Not exactly sure what went wrong in the prev/next spec, but once that is fixed, this is good to go from my perspective.

banukutlu:
> This looks great! Love seeing our usage of components go up. 🚀
> 
> Not exactly sure what went wrong in the prev/next spec, but once that is fixed, this is good to go from my perspective.

Thanks Nick, glad that you like the refactoring here. I will also need to add collapsing for when there are 3+ links for each group.

banukutlu:
This is ready for review again, except a feature spec to test the toggle to view more/less links. I cannot find an example record that has more than 3 links for a type. I added a component test though but would be nice to test the toggling in a feature spec. I will check with Ruth and Maryam to see if we can find a record and get that to included as part of the sample.

nicktiberi:
Here's a record that has 4 external links: https://catalog.libraries.psu.edu/catalog/24798226

app/javascript/psulib_blacklight/styles/_external_links.scss
```diff
@@ -8,3 +8,13 @@
...
+
+// scss-lint:disable QualifyingElement
+button.btn.toggle-external-links.collapsed::before {
+  content: 'View More';
+}
+
+button.btn.toggle-external-links::before {
+  content: 'View Less';
+}
+// scss-lint:enable QualifyingElement
```
nicktiberi, line 20: You can actually remove the `button` prefix on these selectors — it's not actually needed, and then you can also remove the `scss-lint` comments. `.btn` is enough to target these elements.

Your selectors can just be `.btn.toggle-external-links.collapsed::before` and `.btn.toggle-external-links::before`

nicktiberi (commented):
I just have a minor comment about SCSS linting, but overall this looks and works great! I also posted a link to a record with 4 links, so maybe we can pull it in and use it in a feature spec.

app/javascript/psulib_blacklight/styles/_external_links.scss
banukutlu: ahh thanks! will do this change

banukutlu:
> Here's a record that has 4 external links: https://catalog.libraries.psu.edu/catalog/24798226

yeasssss! awesome! you rock thanks 🎉  👏🏻  Let me see if I can get this one added to to sample file.

nicktiberi (approved):

PR #963, opened 2022-03-21
Adds stemming to 'text' fieldType

### LINKED ISSUE #954: Solr not returning stemmed results
keyword: `concentration camp`, format: ` Maps, Atlases, Globes` returns one result

keyword: `concentration camps`, format: ` Maps, Atlases, Globes` returns 6 results.

Those second set of results should be included in the first set too. While unstemmed results should be higher, they should still be in them.

ajkiessl:
There appears to be more than one way to handle this.  I'm not sure which is the best way.

~~I added stemming to the 'text' fieldType, which will impact a lot of the search fields.  Any field defined in the solr schema.xml that has `type="text"` will now use stemming.  I'm not sure if that'll be a problem.  I can scope it down if we think that would be better.~~

Updated this to add another full english text search using stemmer on top of the other full text search that doesn't use stemmer.  Doing this instead of adding the stemmer to all the 'text' fields that I mentioned in the strikethrough above ☝️ .

closes #954

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -19,6 +19,9 @@
     <field name="all_text_timv" type="text" stored="false" indexed="true" multiValued="true" termVectors="true"
            termPositions="true" termOffsets="true"/>
 
+    <field name="all_text_teimv" type="text_en" stored="false" indexed="true" multiValued="true" termVectors="true"
+           termPositions="true" termOffsets="true"/>
+
 
     <!-- NOTE:  not all possible Solr field types are represented in the dynamic fields -->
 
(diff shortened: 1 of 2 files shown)


banukutlu:
@ajkiessl thanks Alex, this solution looks reasonable to me. I am kinda leaning towards to having a preview branch for this change and have a full index in that to let @ruthtillman do some searches and review the search results before we merge. What do you think?

ajkiessl:
@banukutlu Yeah I agree.  I renamed the branch to be a preview and opened a PR.

banukutlu:
> @banukutlu Yeah I agree. I renamed the branch to be a preview and opened a PR.

awesome thanks!

PR #966, opened 2022-03-22
fix for call number browse non matching nearby 

### LINKED ISSUE #964: Call Number Browse error when searching for non-matching query and "You're looking for" is the first result
When searching for a non-matching query that does not have any before items, it throws an error.

eg. search for https://catalog.libraries.psu.edu/browse/call_numbers?classification=lc&nearby=A

banukutlu:
Closes #964


nicktiberi (approved):
LGTM! Tested it locally and works great. 👌

PR #973, opened 2022-03-28
Preview/cjk search

### LINKED ISSUE #950: Traditional and simplified Chinese searching not behaving interchangeably
Just received a report that our traditional and simplified Chinese character results are not behaving interchangeably. This means that if materials are described with simplified Chinese characters but the person searches using traditional Chinese, they will not find the materials. And vice-versa.. The CAT handled this successfully as does WorldCat.

Examples provided were:

```
梵蒂冈图书馆所藏汉籍目录  / 梵蒂岡圖書館所藏漢籍目錄

林则徐全集 / 林則徐全集
```

I can work with our Chinese-language catalogers to get some more examples.

It's a high priority because it affects access to all our Chinese materials. Unfortunately, this was only just now reported because the reporter had not been using the Catalog until the CAT went away. 

CBeer shared in Blacklight Slack that they handle it this way:

https://github.com/sul-dlss/SearchWorks/blob/master/config/solr_configs/schema.xml#L481

and with https://github.com/sul-dlss/CJKFilterUtils 

JRochkind suggested we look for a Solr analyzer. We might also be able to work with Radu on it?

ajkiessl:
This is almost ready to go.  @ruthtillman is still doing some testing.  I basically followed Stanford's old setup using some of the links Ruth provided in #950 .  Looks like their new setup uses a newer version of Solr, so it's not quite like their current code.  It seems to work though. 

I should also note that longer searches like: 中國-台灣問題 : (涉台幹部讀本) 配套資料 : 從對岸看台灣, 國台辦權威文件 seem to blow up Solr on the preview branch.  Idk if this is just from a resource constraint on our preview instances or something else we should be concerned about.

closes #950 

Preview: https://catalog-cjk-search.dev.k8s.libraries.psu.edu


ajkiessl:
Yeah, I can add a test and try to add the fixture.  I'll let you know if I run into any snags.

banukutlu (dismissed):
great work thanks @ajkiessl 👌🏻

banukutlu:
> @banukutlu Adding the fixture broke some of the browsing tests, but looks like it just broke them because adding another record shifted what is returned in the datatables over by one. So, I basically just copied the rspec 'actual' result and pasted it into the expect statements. Let me know if this doesn't seem right, though.
> 
> It looks like codeclimate still isn't satisfied ☹️ .

Alex your updates to specs looks correctly done 👌🏻  The coverage % is a bit better than the previous but still a bit lower but I think that should be fine.

banukutlu:
> > @banukutlu Adding the fixture broke some of the browsing tests, but looks like it just broke them because adding another record shifted what is returned in the datatables over by one. So, I basically just copied the rspec 'actual' result and pasted it into the expect statements. Let me know if this doesn't seem right, though.
> > It looks like codeclimate still isn't satisfied ☹️ .
> 
> Alex your updates to specs looks correctly done 👌🏻 The coverage % is a bit better than the previous but still a bit lower but I think that should be fine, I do not see much in the details on code climate 🤔

OK I compared the test coverage analysis on main and this branch catalog_controller 2 less lines covered, I guess that is the the difference in the total.

banukutlu (approved):
👍🏻

PR #974, opened 2022-03-29
external links toggle feature specs

banukutlu:
Part of #759


spec/features/availability_spec.rb
```diff
@@ -227,7 +227,7 @@
           expect(page).not_to have_selector 'button', text: /View Less/
           expect(page).to have_xpath './/tbody/tr', count: 5
           click_button('View More')
-          sleep 1 # let collapse animation finish and wait for it to re-collapse
+          # sleep 1 # let collapse animation finish and wait for it to re-collapse
```
nicktiberi, line 230: Should we just remove this instead of commenting it out?

spec/rails_helper.rb
```diff
@@ -97,3 +97,6 @@
 Capybara.configure do |config|
   config.javascript_driver = :chrome_headless # This is slower
 end
+
+# Disable CSS animations which slows down tests
+Capybara.disable_animation = true
```
nicktiberi, line 102: Nice! 🚀

spec/features/availability_spec.rb
banukutlu: ahh I thought I removed it 🤦🏻  yes we better remove, will do :) good catch

spec/rails_helper.rb
banukutlu: isn't it? also read that css animations and transitions might be causing flakey tests so I tried disabling them globally, it also let to get rid of those `sleep 1` in the other specs as well.

nicktiberi (approved):
💯

spec/rails_helper.rb
nicktiberi: Mayyybeee this could fix the flaky spec in myaccount? 🤞 👀

banukutlu: yes we should try this over there 🤞🏻

PR #980, opened 2022-04-04
Do not display hathi link if it cannot be accessed

### LINKED ISSUE #968: Remove HathiTrust "Search Inside" Box (leave View Full Text)
Change logic to only display the HathiTrust box if the item is available as full-text in HathiTrust.

The search inside feature is not very useful. It would be more useful instead to see whether the Google Books API has something that would let you search within.

Proposed WDLL

* [ ] HathiTrust box only appears if offering a full view of the item
* [ ] HathiTrust box does not appear for "search inside" item.

ajkiessl:
~~A link to Google Books will be displayed if there is an OCLC number but HathiTrust access is 'deny' or blank.  This might be overkill.  I can remove it if Ruth thinks it's unnecessary.~~

~~Just displaying the Google Books link if an etas item when etas is turned off.~~

No more Google Books stuff.

closes #968


app/components/external_links/hathi_links_component.html.erb
```diff
@@ -1,13 +1,15 @@
...
-                               alt: 'HathiTrust logo',
-                               class: 'mr-3 d-none d-md-inline') + links[:text], links[:url]) %>
-  </p>
-</div>
+<% if links.dig(:open_ht_access) || links.dig(:etas_item) -%>
+  <div class="bs-callout bs-callout-warning">
+    <% if links.dig(:etas_item) %>
+      <h5 class="record-view-only" id="conveying-meaning-to-assistive-technologies">Full Text available online</h5>
+    <% end %>
+    <% if links[:additional_text] %>
```
banukutlu, line 6:
```suggestion
    <% if links[:additional_text] -%>
```
`-` removes the line break for that line

app/components/external_links/hathi_links_component.html.erb
```diff
@@ -1,13 +1,15 @@
...
-  <% end %>
-  <p>
-    <%= link_to(image_pack_tag('media/psulib_blacklight/images/128px-HathiTrust_logo.svg.png',
-                               alt: 'HathiTrust logo',
-                               class: 'mr-3 d-none d-md-inline') + links[:text], links[:url]) %>
-  </p>
-</div>
+<% if links.dig(:open_ht_access) || links.dig(:etas_item) -%>
+  <div class="bs-callout bs-callout-warning">
+    <% if links.dig(:etas_item) %>
```
banukutlu, line 3:
```suggestion
    <% if links.dig(:etas_item) -%>
```
`-` removes the line break for that line

app/models/hathi_links.rb
```diff
@@ -40,4 +41,8 @@ def ht_additional_text
     def etas_item?
       ht_access == 'deny' && Settings&.hathi_etas
     end
+
+    def open_ht_access?
+      @_source['ht_access_ss'] == 'allow'
```
banukutlu, line 46:
```suggestion
      ht_access == 'allow'
```

spec/components/external_links/hathi_links_component.html.erb_spec.rb
```diff
@@ -4,37 +4,69 @@
 
 RSpec.describe ExternalLinks::HathiLinksComponent, type: :component do
   subject(:rendered) do
-    render_inline(described_class.new(links: hathi_links,
-                                      show_links: show_hathi_links))
```
banukutlu, line 8: @ajkiessl I might not be understanding correctly so ignore my comment if your changes are already working the way I am mentionin here: we probably wanna leave the show_hathi_links checks in place, it should be working the same way when show_hathi_links is true and update the spec for when show_hathi_links false.

ajkiessl: I think you're right.  I was a little confused with that at first, but I see this is used in the BaseLinksComponent that HathiLinksComponent inherits from.  I'll change this.

banukutlu (approved):
looks great, thanks so much @ajkiessl 👌🏻

PR #983, opened 2022-04-05
user login

### LINKED ISSUE #836: User Logins
In order to make decisions at the application level we need to support user logins. (#819 ) We can do this via headers set from oauth2 proxy, or via an oauth2 strategy within the application. We've done both of these methods in the past. with the employee-directory app, and with scholarsphere. We'll need an login endpoint that sets a session for a user, ideally we'd get a users group (or role) information, and make decisions based off of the inclusion of that group (or role). 

Since blacklight is largely a public application, user management probably shouldn't bubble up to the user, but we should make it fairly simple to login

whereismyjetpack:
Fixes #836 

Adds ability to support penn state users as well as guest users, right now logins aren't intended for general public use, so we've hidden the 'login' behind a `/login` route. once logged in, we pull user email, and groups from our oauth-proxy, we can use this information to allow admins to do adminy things in the future. right now it drives weather or not mini-profiler is shown 

To use locally, we have to set the HTTP header `X_AUTH_REQUEST_EMAIL` in our browser. I'm using mod-headers for testing 
Screen Shot 2022-04-05 at 4 51 58 PM

We might want to, eventually, support a `/logout` route, and maybe in the future wrap things like `bookmarks` in a `if current_user` block? I'm thinking out loud on those. 

As another PR, I may want to code up an /admin interface, where we can maybe do things like set banners, index records, etc.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -14,17 +14,32 @@ class ApplicationController < ActionController::Base
 
   helper_method :blackcat_config
 
+  def login
+    session[:redirect_url] = home_or_original_path
+    session[:groups] = request.env.fetch(Settings.groups_header, '').split(',')
+    if current_user
+      redirect_to session[:redirect_url] || '/'
+    else
+      redirect_to login_path
+    end
+  end
+
   before_action do
     authorize_profiler
   end
 
   private
 
-    # @note The Matomo ID is used to capture statistics from our production instance. This is very unlikely to change.
-    # We can use this method to limit profiling to only dev and preview instances until we can implement a way to limit
-    # it to particular users and UMGs.
+    def home_or_original_path
+      original_fullpath = request.env.fetch('ORIGINAL_FULLPATH', '/')
+      # prevent redirect loops when user hits /login directly
+      return '/' if original_fullpath == '/login'
+
+      original_fullpath
+    end
+
     def authorize_profiler
-      return if Settings.matomo_id.to_i == 7
+      return unless request.session.fetch(:groups, []).include?(Settings.admin_group)
 
       Rack::MiniProfiler.authorize_request
     end
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -35,7 +35,7 @@ RSpec.describe ApplicationController do
       sign_in user
     end
 
-    it 'does not set a the group' do
+    it 'does not set the group' do
       get :login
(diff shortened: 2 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -14,17 +14,32 @@ class ApplicationController < ActionController::Base
...
+      original_fullpath = request.env.fetch('ORIGINAL_FULLPATH', '/')
+      # prevent redirect loops when user hits /login directly
+      return '/' if original_fullpath == '/login'
+
+      original_fullpath
+    end
+
     def authorize_profiler
-      return if Settings.matomo_id.to_i == 7
+      return unless request.session.fetch(:groups, []).include?(Settings.admin_group)
```
banukutlu, line 42: 😎  
nice that we no longer need to check matomo_id :)

app/controllers/application_controller.rb
```diff
@@ -14,17 +14,32 @@ class ApplicationController < ActionController::Base
...
   before_action do
     authorize_profiler
   end
 
   private
 
-    # @note The Matomo ID is used to capture statistics from our production instance. This is very unlikely to change.
-    # We can use this method to limit profiling to only dev and preview instances until we can implement a way to limit
-    # it to particular users and UMGs.
+    def home_or_original_path
```
banukutlu, line 33: 👌🏻

spec/controllers/application_controller_spec.rb
```diff
@@ -3,26 +3,73 @@
...
+  context 'when given the right headers' do
+    before do
+      @request.env['devise.mapping'] = Devise.mappings[:user]
+      @request.headers[Settings.user_header] = 'user1234@psu.edu'
+      @request.headers[Settings.groups_header] = Settings.admin_group
+      user = User.new(email: 'user1234@psu.edu')
+      sign_in user
+    end
+
+    it 'sets a the group' do
```
banukutlu, line 15:
```suggestion
    it 'sets the group' do
```

spec/controllers/application_controller_spec.rb
```diff
@@ -3,26 +3,73 @@
...
+    end
+  end
+
+  context 'when given the wrong headers' do
+    before do
+      user = User.new(email: 'user1234@psu.edu')
+      sign_in user
+    end
+
+    it 'does not set a the group' do
```
banukutlu, line 38:
```suggestion
    it 'does not set the group' do
```

PR #984, opened 2022-04-06
CGI.escape password

whereismyjetpack:
When a solr password has special characters in it, connections to rsolr were failing, escaping the password fixes this issue. 

rsolr also needs upating to send the password to solr properly. see this PR https://github.com/rsolr/rsolr/pull/232


banukutlu (approved):
lgtm ✔️  thanks for the fix @whereismyjetpack!

PR #986, opened 2022-04-07
Call Number Browse updates to use custom sort plugin - stringdiff

### LINKED ISSUE #961: Call Number Browse updates to use custom sort plugin
### WDLL

- Use [call number sort plugin](https://github.com/psu-libraries/call_number_sort) for call number browse, replace the current terms query.

Requires #960 and Related to psu-libraries/psulib_traject#391

banukutlu:
Closes #961


whereismyjetpack:
the stringdiff plugin has been installed on the production solr nodes. I don't anticipate a problem, but if there is we'll have to revert the config. we might want to do the next deployment during a maintenance window to be extra careful

whereismyjetpack:
stringdiff plugin is now on qa nodes.

nicktiberi:
Finally getting to this. Overall this looks really good to me. Really great work powering through this! 🚀 

I found one issue so far: when you type in a call number that starts with a lowercase letter (e.g. `aBC`), you get an error of `undefined method `first' for nil:NilClass` in `shelf_list.rb`. Maybe we could just uppercase the user-entered call number since I think they are all uppercase anyway?

banukutlu:
> one

great catch, yes upcase makes sense, will do! thanks Nick!

ajkiessl (dismissed):
This all looks great to me.  Nice work!

whereismyjetpack (approved):

app/services/shelf_list.rb
```diff
@@ -20,7 +20,7 @@ def initialize(
     reverse_limit:,
     classification:
   )
-    @query = query
+    @query = query.upcase
```
banukutlu, line 23: nice, thanks for catching this and adding a fix!

nicktiberi (approved):
Great fixes, @banukutlu! Let's gooo! 🚀

PR #987, opened 2022-04-07
Use I18n.t! to raise errors if translations are missing in tests

### LINKED ISSUE #775: Use I18n.t! in tests
Using `I18n.t!` in our test suite will raise errors if any translations are missing.

nicktiberi:
Closes #775.

If any of the translation keys used in our tests get deleted, the tests reading those keys will fail and let us know something is wrong.


banukutlu (approved):
👍🏻

PR #989, opened 2022-04-07
Preview/userlogin

whereismyjetpack:
If i user visits the catalog without logging in, `devise-guests` still runs them through the login routine, and they get a `:guest_user_id` object in their session

If they later visit the homepage, or login routine after sending the appropriate headers, they have both an `:id` and a `:guest_user_id` object in their session

Here, upon successful login, we remove the `:guest_user_id` object. this fixes the ``undefined method `relation_delegate_class' for Search:Class`` bug


whereismyjetpack:
@ajkiessl i was able to test and reproduce this on my local, but would you mind verifying for your local ?

ajkiessl:
It works!

banukutlu:
worked on my local as well 🎉

banukutlu (approved):
✔️  it works! great fix thanks @whereismyjetpack

PR #993, opened 2022-04-11
remove master.key requirement

### LINKED ISSUE #722: Use vault
When setting up a new install, you need the `config/master.key` to decrypt credentials, even for local development. An alternative solution could be to move the management of credentials to Vault. This would, we think, remove the need for the local encrypted master key, and the various secrets and database passwords would be manged by vault for each server deployment.

@whereismyjetpack can weigh-in on the this, particularly regarding the feasibility of this.

whereismyjetpack:
In kubernetes, we set the `MYSQL_*` variables as part of the deployment, on linux we can set them in settings.local

for `secret_key_base` we set this as part of the deploy users' env, in k8s we set this from vault 

TODO 

- [x] set the `Setting.mysql.*` values in prod 
- [x] set the `SECRET_KEY_BASE` env variable in prod 

Fixes #722


banukutlu (approved):
LGTM 👍🏻  thanks @whereismyjetpack

PR #995, opened 2022-04-15
Google Books Preview Link

ajkiessl:
Displays Google Books preview link on show page if a preview or full view exists in Google Books.

Screen Shot 2022-04-15 at 12 44 58 PM

I mentioned in standup that the delays in the ajax calls were preventing VCR from detecting the requests to Google, so I had to add sleep statements.  I think the requests to Google were actually the slow part, so once the VCR cassettes were registered I was able to remove the sleep statements and the tests still work.  Also, for some reason a couple of the `Availability` tests were not registering the Google requests properly with VCR, so I had to manually add them by copying from the cassettes that did work.


app/components/external_links/google_preview_link_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="bs-callout bs-callout-info" id="google-preview" style="display:none;" data="<%= search_item %>">
```
nicktiberi, line 1: It might be a little cleaner/more clear if you specified the data attribute as something like `data-search-item`. Then in `google_preview.js` you could access it via `element.dataset.searchItem` instead of `element.attributes.data.value.toString()`. Not a dealbreaker though; just a style thing.

app/javascript/psulib_blacklight/javascripts/google_preview.js
```diff
@@ -0,0 +1,27 @@
+$(document).ready(() => {
+  const element = document.getElementById('google-preview');
+  const googlePreviewUrl = '/preview/google-preview-data';
+  const googlePreviewUrlQuery = `${googlePreviewUrl}/?search_item=${element.attributes.data.value.toString()}`;
+
+  if (element) {
+    $.ajax({
+      url: googlePreviewUrlQuery,
+      success(response) {
+        for (const key in response) {
```
nicktiberi, line 10: Does this need to be a loop? Do we think there may potentially be more than one response here? Or that the response key may not match up exactly with the `search_item` param we are passing in?

I'm fine if the answer to any of these questions is yes; just want to make sure I'm understanding the rationale.

I was thinking we could just do

```
if (response[search_item])
  //do stuff
```

but maybe that isn't reliable enough and your approach would catch the scenarios above.

nicktiberi (approved):
A couple minor comments, but this looks good to me overall. Nothing noted is a dealbreaker IMO. Nice work! 📚

nicktiberi:
One other quick question — where did we get the Google Books button? I ask because it is pretty low-res. If we could find a higher-res version, that would be awesome, but if not, it's fine to roll with what we have.

ajkiessl:
> One other quick question — where did we get the Google Books button image? I ask because it is pretty low-res. If we could find a higher-res version, that would be awesome, but if not, it's fine to roll with what we have.

I got it from here: https://www.google.com/intl/en/googlebooks/images/gbs_preview_button1.gif

This is the image Google recommends to use.

nicktiberi (approved):
LGTM! Thanks, @ajkiessl! 📖

ajkiessl:
`google_preview.js` now uses the search item as the key to grab the response from the API.  I added some logic in there to loop through all the values if nothing could be retrieved using the search item.  I also added a test for this.

Issue #55, opened 2018-09-21
Google Books integration / preview


banukutlu:
Determine whether we integrate Google Books (or Open Library), including its preview links.

Example 1 partial-text: Title: `Colonial interiors. second series` OCLC # `2628707` and Google Books link goes to https://books.google.com/books?id=RvQ8AAAAYAAJ&source=gbs_ViewAPI

Example 2 full-text: Title: `Farmers' institutes in Pennsylvania to be held under the auspices of the Pennsylvania Department of Agriculture.` OCLC # `40314611` and Google Books link goes to: https://books.google.com/books?id=1Qg2AQAAMAAJ&printsec=frontcover&source=gbs_ViewAPI#v=onepage&q&f=false

The examples and screenshots are the same as for Issue #54 and example 2 on Google Books is our copy.

![hathitrustexample1](https://user-images.githubusercontent.com/6477867/46040583-b662c880-c0de-11e8-851e-3027b222ae93.png)

![hathitrustexample2](https://user-images.githubusercontent.com/6477867/46040585-b9f64f80-c0de-11e8-8a78-72488e0f6c3a.png)

We could also try the OpenLibrary API which is allows for search inside -- or query that first. https://openlibrary.org/dev/docs/api/books

banukutlu:
In GitLab by @rkt6 on Jun 21, 2018, 15:40

changed the description

ruthtillman:
Note: Google Books, we'll want to test whether the book can actually be searched inside/previewed or not.

Do we want to do some kind of assessment test of something like OpenLibrary to see how much we can get from there? I won't be meeting with Ellysa until the 15th however to talk about how usable that is

ajkiessl:
https://developers.google.com/books/docs/v1/using#download-format

ajkiessl:
Ok, so it looks like Google has a concise API used specifically for linking to previews: https://developers.google.com/books/docs/dynamic-links?csw=1

You can search using these params:

ISBN
    &bibkeys=ISBN:0451526538 (The API supports both ISBN 10 and 13.)
OCLC
    &bibkeys=OCLC:36792831
LCCN
    &bibkeys=LCCN:96072233

The "preview" element in the returned JSONwill tell us if a preview exists.  This value can be "full", "partial", or "noview".
The preview url (if one exists) can be pulled from the "preview_url" element in the JSON.

 It does not require an API key.

ruthtillman:
[...]

PR #996, opened 2022-04-19
55 additions

ajkiessl:
Some fixes/additions to #55 + #995.  

- Do not display Google Preview link if the record is 'Free to Read' or has a HathiTrust link
- Wording changes for Google Preview link ~~(waiting for Binky's recommendations)~~


app/components/external_links/google_preview_link_component.rb
```diff
@@ -30,6 +32,17 @@ def isbn
         document[:document]['isbn_valid_ssm']&.first
       end
 
+      def free_to_read?
+        access_facet = document[:document]['access_facet']
+        return false if access_facet.blank?
```
nicktiberi, line 37: Do we need this line? I think we really only care about the check on line 39.

ajkiessl: It should probably be changed to use `.nil?`.  Sometimes the access_facet is nil and nil will raise an error when `.include?` is sent to it.  I used `.blank?` just in case there's an empty string returned, but that's probably not possible since this field will either be an array or nil.

banukutlu (changes requested):
[psulcat_test-journal](https://github.com/psu-libraries/psulib_blacklight/pull/996/files#diff-27506025005f6797894d139a860dbcb9001b8fd41253369f957c9d4788435be8) should this be ignored?

ajkiessl:
> These changes looks great so far 👍🏻 only one comment: [psulcat_test-journal](https://github.com/psu-libraries/psulib_blacklight/pull/996/files#diff-27506025005f6797894d139a860dbcb9001b8fd41253369f957c9d4788435be8) should this be ignored?

I think it should just be deleted.  It was added with the initial Google Previews PR: https://github.com/psu-libraries/psulib_blacklight/commits/main/psulcat_test-journal .  I have no idea what it's for our how it got there, so I'm thinking it should be deleted.

banukutlu:
> > These changes looks great so far 👍🏻 only one comment: [psulcat_test-journal](https://github.com/psu-libraries/psulib_blacklight/pull/996/files#diff-27506025005f6797894d139a860dbcb9001b8fd41253369f957c9d4788435be8) should this be ignored?
> 
> I think it should just be deleted. It was added with the initial Google Previews PR: https://github.com/psu-libraries/psulib_blacklight/commits/main/psulcat_test-journal . I have no idea what it's for or how it got there, so I'm thinking it should be deleted.

yeah, would you mind deleting it with this PR.  thanks!

nicktiberi (approved):
Looks great! 🥇

PR #999, opened 2022-04-21
Availability: Change Summary Holdings h6 elements to divs

### LINKED ISSUE #991: Availability: Make Summary Holdings Headers normal text
Per accessibility testing with Michelle/Ruth/Binky, the "Holdings Summary" headings in the blue sections of the tables are confusing, and it's frustrating to have so many extra heading tags in the document because it makes navigation more difficult. It would be better to just have that text be a `<span>` or something with a `h6` style.

From:

```
<h6>{locationName}: Holdings Summary</h6>
```

To:
```
<span className="h6">{locationName}: Holdings Summary</span>
```

nicktiberi:
Closes #991.

In screen reader testing with Michelle, we found out that having too many header tags on the page actually made navigation more difficult and confusing. Replacing these headers with `div`s with an `h6` class provides the same visual appearance but makes screen reader navigation easier.

### DIFF
--- a/app/javascript/availability/components/summary_holdings.jsx
+++ b/app/javascript/availability/components/summary_holdings.jsx
@@ -21,7 +21,7 @@ const SummaryHoldings = ({ summaryHoldings }) => {
 
           return (
             <React.Fragment key={locationID}>
-              <h6>{locationName}: Holdings Summary</h6>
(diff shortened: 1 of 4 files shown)


ajkiessl (approved):
This all looks good to me!

banukutlu (approved):
lgtm 👍🏻

PR #1001, opened 2022-04-22
Hotfix- Google preview link was still showing when Hathi link was present

ajkiessl:
I had the document parsing wrong for grabbing the hathi trust access field: 'ht_access_ss'.  I think `document[:document]['ht_access_ss']` should work.  I deleted my setup for getting the hathi stuff locally so I'm having trouble confirming this, though.


banukutlu (approved):
good catch and fix 👍🏻  

the test is passing and we can test on qa as well, no worries for local imo ;)

PR #1005, opened 2022-04-28
Fix "Ask a Librarian" widget not showing on bookmarks page.

### LINKED ISSUE #608: Ask a librarian does not show on /bookmarks when there are bookmarks
works when there are no bookmarks

ajkiessl:
Remove turbolinks stuff and Rails.start in application.js.  Moved all js implementations in application.js to trigger on DOM load.

Preview is here: https://catalog-ask-librarian.dev.k8s.libraries.psu.edu

These changes don't seem to have broken anything, and the tests are passing, so I think the TurboLinks and Rails js stuff is already running with Rails and doesn't need to be handled with node/webpacker.

closes #608

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -7,8 +7,6 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
-import Rails from 'rails-ujs';
-import Turbolinks from 'turbolinks';
 import availability from '../availability';
 import bookCovers from '../book_covers';
 import search from '../search';
@@ -19,11 +17,6 @@ import '../blacklight_overrides';
 
 require.context('../psulib_blacklight/images/', true);
 document.addEventListener('DOMContentLoaded', () => {
-  Rails.start();
-  Turbolinks.start();
-});
-
-document.addEventListener('turbolinks:load', () => {
   availability.loadAvailability();
   bookCovers.start();
   search.autoPlaceholder();
--- a/package.json
+++ b/package.json
@@ -19,8 +19,7 @@
     "prop-types": "^15.7.2",
     "rails-ujs": "^5.2.6",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "turbolinks": "^5.2.0"
+    "react-dom": "^17.0.2"
   },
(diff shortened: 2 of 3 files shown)


ajkiessl:
Also, looks like this commit added in the turbolinks stuff: https://github.com/psu-libraries/psulib_blacklight/commit/8574b27fddf447429b04fa9acd75f1b4a97f35ea

ajkiessl:
And this one added the `Rails.start`: https://github.com/psu-libraries/psulib_blacklight/commit/f1eb95b39d3b415412ead27f39d00e4e7f579348

nicktiberi (approved):
Nice! I wonder if the inclusion of Rails/Turbolinks start was inadvertently firing double events which was confusing the Ask a Librarian code. 🤔 

FYI, looks like this is where Ask a Librarian gets loaded: https://github.com/psu-libraries/psulib_blacklight/blob/5e2ecf1230c03c11691753b4a21488ce16fc4b3c/app/views/layouts/blacklight/base.html.erb#L35

I guess at some point this was handled a different way based on those commits you linked to, @ajkiessl. That code references `executeAskALibrarian` which isn't around anymore.

ajkiessl:
> Nice! I wonder if the inclusion of Rails/Turbolinks start was inadvertently firing double events which was confusing the Ask a Librarian code. 🤔
> 
> FYI, looks like this is where Ask a Librarian gets loaded:
> 
> https://github.com/psu-libraries/psulib_blacklight/blob/5e2ecf1230c03c11691753b4a21488ce16fc4b3c/app/views/layouts/blacklight/base.html.erb#L35
> 
> I guess at some point this was handled a different way based on those commits you linked to, @ajkiessl. That code references `executeAskALibrarian` which isn't around anymore.

Yeah, I'm still not 100% sure how the Rails/Turbolinks stuff was affecting this since the javascript for the "Ask" widget is embedded in the html.

nicktiberi:
> > Thanks @ajkiessl, I do not see the homepage facets loaded on the preview, guess it might be related to this change? Could you also check that?
> 
> Not too sure why they are not showing in the preview. It works locally. I can try reindexing on the preview.

It might be the rails cache. You can try clearing it with `Rails.cache.clear`.

ajkiessl:
> Rails.cache.clear

Yup, that fixed it!

PR #1011, opened 2022-05-04
return plain text when format is not html

### LINKED ISSUE #1010: render all 404
Presently, we only render 404 for `handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]` -- when a request comes in for something else, like `ids.txt` we throw a template error instead of a 404

whereismyjetpack:
When a user navigates to a non html path, previously the application would throw a 500 template not found error, instead of a 404. for non html requests, we fall back to rendering plain text, with status of 404. 

Fixes #1010

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -2,7 +2,10 @@
 
 class ErrorsController < ApplicationController
   def not_found
-    render status: 404
+    respond_to do |format|
+      format.html { render status: 404 }
+      format.all { render plain: 'not found', status: 404 }
+    end
   end
 
   def internal_server_error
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
LGTM 👍🏻  thanks @whereismyjetpack 🍻

banukutlu:
just find out that we actually have specs for this one, I totally forgot about it, it is a request spec here https://github.com/psu-libraries/psulib_blacklight/blob/main/spec/requests/errors_spec.rb

whereismyjetpack:
writing up a spec shouldn't be too hard...

_3 days later_

whereismyjetpack:
I've added a request spec to ensure this behavior persists.

banukutlu (approved):
looks great to me, thanks @whereismyjetpack!

PR #1014, opened 2022-05-11
Availability: add more component tests + minor cleanups

nicktiberi:
Some more tests could still be added to the `LocationInfo` component to get the coverage up, but that would require mocking `AeonLink` and `IllLink` inside the location info tests. For some reason I couldn't get that to work, so I abandoned for now.


banukutlu (approved):
lgtm thanks @nicktiberi! 🍻

PR #1015, opened 2022-05-17
fix for call number browse next button 

banukutlu:
next button was selecting the last item before slicing to length, which was causing paging to jump further than we want


nicktiberi (approved):
👍

PR #1060, opened 2022-10-24
change illiad isbn

### LINKED ISSUE #1054: Substitute ISBN field for ILLIAD
Request from ILL dept -- we are currently sending isbn_ssm values which may include things like "(hardcover)" or "pbk" in addition to the ISBN itself. This can cause issues with their automated processes. We also index isbn_valid_ssm, which is only the ISBN itself. Can we substitute isbn_valid_ssm here https://github.com/psu-libraries/psulib_blacklight/blob/main/app/javascript/availability/components/ill_link.jsx#L43-L44 and anywhere else it might be passed?

Also -- would it be possible to send just the first ISBN from the array, they greatly prefer one. So isbn_valid_ssm[0]?

Sample records:

- https://catalog.libraries.psu.edu/catalog/22083361/raw.json 
- https://catalog.libraries.psu.edu/catalog/37419096/raw.json

## WDDLL

- [ ] We're sending ISBNs without extra textual data/qualifying data
- [ ] We're sending just one ISBN.

Smullz622:
changed isbn_ssm field to isbn_valid_ssm[0] in order to return a single numeric isbn value
fixes #1054

### DIFF
--- a/app/javascript/availability/components/ill_link.jsx
+++ b/app/javascript/availability/components/ill_link.jsx
@@ -40,7 +40,7 @@ const IllLink = ({ holding }) => {
           if (linkType === 'archival-thesis') {
             illUrl += '&Form=20&Genre=GenericRequestThesisDigitization';
           } else {
-            const ISBN = data.isbn_ssm ? data.isbn_ssm : '';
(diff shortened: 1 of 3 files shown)


banukutlu (approved):
looks great 👌🏻

PR #1067, opened 2022-10-28
233 feature tests

Smullz622:
Here's what I've done so far

Advanced Search:
- Single term is pretty much done. I got stuck on the publication date range context so that will need attention.
- I didn't get to vernacular search (needs updated search terms) or MARC view

Basic Search (Title, author, subject)
- Framework is complete for the title & author tests. Actual search terms/catkeys need updated once the acceptance test wiki is updated
- Subject tests are completed (I searched for data to use and updated the acceptance test wiki accordingly. @ruthtillman could you please confirm they are good ones to use?)

Part of #233


banukutlu:
@Smullz622 what you have so far looks good to me 👍🏻 just one minor change: can we use `Single Item` instead of `Individual Record`? in filenames and descriptions? Just to keep it consistent with our docs https://github.com/psu-libraries/psulib_blacklight/wiki/Blacklight-Catalog-Fields#single-item-view-labels

Smullz622:
@banukutlu will do! But the structure look good though? Are there any scenarios that I'm missing?

banukutlu:
> @banukutlu will do! But the structure looks good though? Are there any scenarios that I'm missing?

yup I think you have most of it covered, I could not think of another case. We can always add later as we find out more edge cases. What you have here covers nicely 👍🏻

spec/features/advanced_search_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+          click_button 'advanced-search-submit'
+        end
+
+        it 'results include expected CAT keys' do
+          expect(page).to have_selector 'article[data-document-id="17280643"]'
+        end
+      end
+    end
+
+    # should this be in another location?
```
ajkiessl, line 164: It may be a good idea to move this somewhere else since it isn't really related to advanced search.  Would it be helpful to also test clicking the link and testing that the marc record shows?  Not sure if that's being tested anywhere already.

ajkiessl (commented):
This all looks good to me.  I just have one comment.

ajkiessl (approved):
lgtm!

banukutlu (approved):
lgtm 👌🏻  thanks Sarah!

spec/features/advanced_search_spec.rb
banukutlu: agree with Alex, thanks for your review, good point! and thanks Sarah for refactoring that part.

PR #1069, opened 2022-10-31
sidekiq is not used

whereismyjetpack:
In testing with `green` i noticed that we are not using sidekiq in the catalog, but we are calling it in the datadog initializer, this is causing some unneeded output in the prod environment


banukutlu (approved):

banukutlu:
good catch, thanks @whereismyjetpack

PR #1070, opened 2022-10-31
'Ask a Librarian' widget to show on bookmarks

### LINKED ISSUE #608: Ask a librarian does not show on /bookmarks when there are bookmarks

ajkiessl:
Afaict these changes have the widget working and the availability stuff is unaffected.  You can check it out here: https://catalog-ask.dev.k8s.libraries.psu.edu

Not sure what was causing the issues with availability before.

closes #608

### DIFF
--- a/spec/features/bookmarks_spec.rb
+++ b/spec/features/bookmarks_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Bookmarks', type: :feature do
+  before do
+    stub_request(:any, /hathitrust/).to_return(status: 200, body: '{}', headers: {})
+  end
+
+  it 'Adds, removes, re-adds, and views a bookmarked record', js: true do
+    visit '/?utf8=✓&search_field=all_fields&q=Ethical+and+Social+Issues+in+the+Information+Age+AND+9783319707129'
+    bookmark_buttons = find_all('.toggle-bookmark')
+    bookmark_buttons.first.click
+    sleep 1
+    expect(page).to have_content 'In Bookmarks'
+    bookmark_buttons.first.click
+    sleep 1
+    expect(page).not_to have_content 'In Bookmarks'
+    bookmark_buttons.first.click
+    sleep 1
+    click_link 'Bookmarks'
+    expect(page).to have_css :h1, text: 'Bookmarks'
+    expect(page).to have_content '1 entry found'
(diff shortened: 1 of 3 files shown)


banukutlu:
@ajkiessl These changes were not affecting availability, sorry if this was a misunderstanding, the issue is with bookmarks. For example, when I try to bookmark a record on search results and click again to remove bookmark, I get an error as below. Similarly, if a record is already bookmarked, I remove the bookmark and click again to bookmark, I am getting an error again. It seems it is happening when clicked for the same record consecutively. Same behavior on bookmarks page as well. 

Screen Shot 2022-11-02 at 8 36 41 PM

banukutlu (changes requested):
@ajkiessl These changes were not affecting availability, sorry if this was a misunderstanding, the issue is with bookmarks. For example, when I try to bookmark a record on search results and click again to remove bookmark, I get an error as below. Similarly, if a record is already bookmarked, I remove the bookmark and click again to bookmark, I am getting an error again. It looks like it is happening when clicked for the same record consecutively. Same behavior on bookmarks page as well. It is working on qa and catalog. 

Screen Shot 2022-11-02 at 8 36 41 PM

ajkiessl:
Looks like adding `rails-ujs` back in fixed this.  The CSRF protection was blocking subsequent ajax calls.  Not too sure _why_ `rails-ujs` fixes this, but it does.  

I added a test for adding, removing and re-adding a bookmark.  It won't actually catch this issue since it looks like the CSRF protection doesn't block these requests in development and test.  I guess it doesn't hurt to have the test though.

ajkiessl:
> I added a test for adding, removing and re-adding a bookmark. It won't actually catch this issue since it looks like the CSRF protection doesn't block these requests in development and test. I guess it doesn't hurt to have the test though.

I found a way to turn on CSRF protection in the test.

spec/features/bookmarks_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Bookmarks', type: :feature do
+  before do
+    # In Rails 6, CSRF protection is turned off by default in test env
+    # Turning it on here to test that it doesn't break Bookmarks
+    ActionController::Base.allow_forgery_protection = true
```
banukutlu, line 9: 👌🏻

banukutlu (approved):
looks great, thanks @ajkiessl 🙌🏻

PR #1077, opened 2022-11-09
Sort links

### LINKED ISSUE #1007: Label Finding Aids separately from other Supplemental Links
Some of our supplemental links are to what are called "finding aids," detailed documents which allow a patron to learn more about and request archival materials that are more briefly described in the catalog. These are all already being indexed as `suppl_links_struct` but would benefit from a clearer name to let the person know it's not actually supplemental/related material, it's a place to get more information about the materials described by the record. This is important for highlighting Penn State's unique materials/archival collections.

## WDDLL

- [ ] Links to our finding aids have the label "Special Collections Materials" (this is my tbd header)
- [ ] Should list before any other Related Resources

## Finding Aid URLs

We used to have more ways of linking to finding aids, but all those should be part of our cataloging side cleanup, so the ticket was revised just to match our ARKs. Our base URLs should be the following:

```
http(s)://n2t.net/ark:/42409/fa8[rest of url]
```

So we can use:

```
*ark:/42409/fa8*
```

to be safer.

### Catkeys

These catkeys represent records with ARKs.

- 9182662
- 374137
- 4732299

ajkiessl:
Sorting links so the 'Special Collections Materials' link is at the top.  Depends on https://github.com/psu-libraries/psulib_traject/pull/435.

Closes #1007

### DIFF
--- a/app/components/external_links/base_links_component.rb
+++ b/app/components/external_links/base_links_component.rb
@@ -36,9 +36,8 @@ module ExternalLinks
       def sorted_links(links_arg)
         return links_arg unless links_arg.is_a?(Array)
 
-        links_arg.sort_by{ |l| l['text'] == 'Special Collections Materials' ? 0 : 1 }
+        links_arg.sort_by { |l| l['text'] == 'Special Collections Materials' ? 0 : 1 }
       end
-
 
       def show_links?
         !!@show_links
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
👌🏻  looks great! thanks @ajkiessl

PR #1084, opened 2022-11-21
Separates Special Collections Materials from other Related Resources

ajkiessl:
Some more context at the bottom of the thread here: #1007

We basically want anything with an ARK url to display in its own box indicating it's Special Collections Materials.  This is how it will look:

Screen Shot 2022-11-21 at 1 44 48 PM

Other related resources will still be in their Related Resources box. 

I modified the PsuDigitalCollectionsLinks component to handle digital and special collections since they are expected to display almost the same way.  That class is now a bit more polymorphic and I renamed it to PsuCollectionsLinks so it fits both types.

Depends on https://github.com/psu-libraries/psulib_traject/pull/440


banukutlu (approved):
lgtm 👌🏻  thank!

PR #1087, opened 2022-11-28
Initialize Tooltips in Availability "View More"

### LINKED ISSUE #1072: Public note tooltip behavior
Not critical, but maybe we could brainstorm how the public note tooltips work in the item display. Because it's not immediately apparent that they're tooltips, one person brought her mouse over it and, not seeing anything happen, clicked, and didn't se it. Based on my own experimentation, I'm guessing it took more time than she expected for it appear but then clicking closed it. I also got ??? results on mobile. What we have now technically works, but it's something we might revisit.

She was using Chrome.

example record: https://catalog.libraries.psu.edu/catalog/2169033 (you have to see the copies in More to see the one with the note)

ajkiessl:
closes #1072 

The problem here was that the tooltips within the "View More" records in the availability section were not being initialized on page load because they do not exist until the "View More" button is clicked.  I added a function to the availability code to reinitialize the tooltips when the "View More" button is clicked.  I tried placing the reinitialize code in several different places.  I could only get it to work within the `viewMore` constant.  There may be a better place to put this, but it seems to work as is.


banukutlu (approved):
this makes so much sense, looks great! thanks @ajkiessl 🚢

PR #1088, opened 2022-11-30
Handle availability holdings above 1000

### LINKED ISSUE #1083: Brainstorm loading strategies for records with thousands of holdings
Some of our records have thousands of holdings, so even doing something like loading every 100 is going to be a lot of clicks. Sample record (has about 2500 map sheets attached): https://catalog.libraries.psu.edu/catalog/1665061

How might we improve this experience?

One thought: is there something we could do like -- after a person has loaded the first 300 records we start loading in 500s because at this point they're aware it's an enormous record and perhaps they'll tolerate slower load times? How would this affect load time?

Or is there a way to see how many items up front and parcel out how many we load based on rough number ranges? (while not coding overkill)

ajkiessl:
Added code to Availability to bump pageSize to 500 if there are more than 1000 holdings.

closes #1083 

Preview: https://catalog-long-avail.dev.k8s.libraries.psu.edu/catalog/1665097

### DIFF
--- a/app/javascript/availability/components/availability.jsx
+++ b/app/javascript/availability/components/availability.jsx
@@ -8,8 +8,11 @@ import ViewMoreButton from './view_more_button';
 const Availability = ({ structuredHoldings, summaryHoldings }) => (
   <>
     {structuredHoldings.map((element, index) => {
+      const totalHoldingsCount = structuredHoldings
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
looks good to me 👍🏻  thanks Alex!

PR #1095, opened 2022-12-05
Libraries locations

ruthtillman:
Found & fixed typo as well. This file has the core BTAA changes that the annex needs deployed in order to start sending stuff to the BTAA location.


banukutlu (approved):
👍🏻

PR #1108, opened 2022-12-21
Sort by purchase date

### LINKED ISSUE #1025: Sort results by "newest-to-oldest" using record number
We've had periodic requests to order search results by newest-to-oldest purchase date (because sometimes people buy new things that were published a while back, e.g. new copies of Dune). While this wouldn't really apply well to e-resources since we're constantly loading/reloading them, it might actually be beneficial for physical materials.

There isn't a great way to do this with MARC data but we could fudge it by ordering results by record number in descending order. e.g. 29835, 3924, 395, 23.

Is this something Solr would do at all well? It's more of a nice-to-have than a priority, but recording it as a frequent request.

Smullz622:
Fixes #1025 

Sorts results from newest to oldest using catkey as a way to determine purchase date. The Solr schema was adjusted, so we will need to reindex.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -488,7 +488,7 @@ class CatalogController < ApplicationController
     config.add_sort_field 'score desc, pub_date_itsi desc, title_sort asc', label: 'relevance'
     config.add_sort_field 'pub_date_itsi desc, title_sort asc', label: 'year'
     config.add_sort_field 'title_sort asc, pub_date_itsi desc', label: 'title'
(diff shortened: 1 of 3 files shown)


ajkiessl (approved):
This all looks good to me.  The only question/comment I have is: are the tests in sort_by_spec.rb testing that the sorting functionality is actually sorting things?  Or is it just testing selecting the sorting method from the dropdown, that it then does something (without breaking), and that it displays something?  Idk if it is worth actually testing that the returned contents have been sorted in the manner we expect?  The sorting functionality is kind of built in to blacklight, so it may not need to be tested to that granularity.

banukutlu (approved):
👍🏻

PR #1110, opened 2023-01-03
Refactored AeonLink and IllLink into SpecialRequestLink

### LINKED ISSUE #869: Combine AeonLink/IllLink into SpecialRequestLink
Since the `<AeonLink />` and `<IllLink />` React components are basically the same except the URL and label, consider combining them into a single `<SpecialRequestLink />` component to avoid duplicating so much logic.

ajkiessl:
Got the two 100+ line classes refactored down to one ~150 line class.  The AeonLink class took an argument (locationText) that IllLink did not take, so I used that as the condition to determine whether the SpecialRequestLink is an Aeon or Ill link.  It's a bit flimsy, but it seems to work.  I'm open to doing this a different way if someone thinks I should.  I didn't do much refactoring with the tests since I didn't see much to be shared.  I basically just combined the tests for the AeonLink and IllLink classes into one test file.

closes #869


ajkiessl:
Going to try to fix the CodeClimate issues

ajkiessl:
I don't think I'll be able to satisfy CodeClimate on it's last issue.  I think in order to reduce the complexity, I'd have to start abstracting things into separate classes.  I think the AeonLinkUrl and IllLinkUrl and their associated functions could be pulled into their own classes, but I don't think I should do anymore right now since I've already changed a lot.

ajkiessl:
Eh, I'll give it a try.

ajkiessl:
I think I made it worse by refactoring more, so I'll leave it like this.  It is ready for review.

banukutlu (approved):
lgtm 👍🏻  thanks @ajkiessl

nicktiberi:
GitHub notified me about this haha 🙂 Nice work! 🚀

PR #1112, opened 2023-01-04
browse by title linking updated

### LINKED ISSUE #889: Need better linking to title search
This may be solved by @awead moving out the controllers and adding title_sort there? But right now when you search for something like American Marriage https://catalog.qa.k8s.libraries.psu.edu/browse/titles?prefix=american+marriage you get some weird things - try first 2 results.

Smullz622:
Fixes #889

### DIFF
--- a/app/views/browse/_titles.html.erb
+++ b/app/views/browse/_titles.html.erb
@@ -12,7 +12,7 @@
   <tbody>
     <% title_list.entries.map do |item| %>
       <tr>
-        <td><%= link_to item.value, search_catalog_path("f[#{title_list.field}][]" => item.value) %></td>
+        <td><%= link_to item.value, search_catalog_path("f[title_sort][]" => item.value) %></td>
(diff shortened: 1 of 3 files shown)


app/views/browse/_titles.html.erb
```diff
@@ -12,7 +12,7 @@
   <tbody>
     <% title_list.entries.map do |item| %>
       <tr>
-        <td><%= link_to item.value, search_catalog_path('search_field' => 'title', 'q' => item.value) %></td>
+        <td><%= link_to item.value, search_catalog_path("f[#{title_list.field}][]" => item.value) %></td>
```
banukutlu, line 15: Do we need `#{title_list.field}` here? We only need title_facet, right?

Smullz622: yep, I can change it

banukutlu (approved):

PR #1116, opened 2023-01-05
No-recalls "I want It" button updates

### LINKED ISSUE #609: No Recalls
To support the strategic plan's no-recalls work, develop logic for the "I Want It" button which directs users to a page which places the hold via ILLiad vs Symphony.

If the item has holdable:true, then we need to test:

 1. Are all items in ILL locations aka: https://github.com/psu-libraries/psulib_blacklight/blob/main/app/javascript/availability/libraries_locations.json#L780-L794
    - yes: trigger No Recalls
    - no: go to 2
2. Are items that are NOT checked out in non-holdable locations:
    - yes: trigger No Recalls
    - no: allow normal hold placement through Symphony.

We will need to figure out volumetrics.

Fuller logic and use cases are spelled out in the [No Recalls Logic](https://github.com/psu-libraries/psulib_blacklight/wiki/No-Recalls-Logic) page in our Wiki. Includes sample catkeys for the possible use cases and info on where to find copies of the WebServices results for those sample records (since checking in and out items will alter whether these are good samples, we can use static results/manipulate our own data on stage).

## WDDLL

- [x] If a record _normally_ contains holdable items but none can be checked out at this time, a user clicking "I Want It" goes to a page on My Account which places the hold in ILLiad vs. in the Catalog.
- [x] If the above but the user is a community borrower/PA borrower, they still get redirected to normal My Account holds page, where they will place a non-recalling hold.

banukutlu:
"I Want It"  button logic update for no-recalls so that it directs users to myaccount's ILLiad place loan page if
- all items are in ILL locations, or
- all items that are NOT in ILL Locations in non-holdable locations
- otherwise, place hold through Symphony
 
More info here: https://github.com/psu-libraries/psulib_blacklight/wiki/No-Recalls-Logic 

Closes #609


config/settings.yml
```diff
@@ -1,4 +1,5 @@
 hold_button_url: https://myaccount.qa.k8s.libraries.psu.edu/holds/new?catkey=
+no_recall_button_url: https://myaccount.qa.k8s.libraries.psu.edu/no_recalls/new?catkey=
```
whereismyjetpack, line 2: I wonder if this could be refactored into `my_account_url` and have the code do something like `#{Settings.my_account_url}/holds/new?catkey=`

is no_recall `/ill/new?catkey=`?

whereismyjetpack (changes requested):

config/settings.yml
banukutlu: yeah that makes sense, I can make that change.

and yup no-recall is `/ill/new?catkey=?`

app/javascript/availability/index.js
```diff
@@ -484,8 +491,7 @@ const availability = {
     return (
       ['UP-MICRO'].includes(holding.libraryID) &&
       holding.homeLocationID !== 'THESIS-NML' &&
-      holding.itemTypeID === 'MICROFORM' &&
-      !(holding.locationID in availability.illiadLocations)
+      holding.itemTypeID === 'MICROFORM'
```
ajkiessl, line 494: I think the change here is causing the test to fail.  Do we want to handle all MICROFORM links the same now, and not differentiate between ILL and non-ILL locations?  If so we'll have to remove some tests: https://github.com/psu-libraries/psulib_blacklight/blob/main/spec/javascript/availability/components/special_request_link.jsx.spec.js#L221

banukutlu: I am assuming we will eventually not link to ILL from the locations since this no recall change is updating I want It button to go to myaccount's ILL page. But I will put this line back for the initial pass of this PR and have Ruth review things and give us feedback re linking the ILL locations or not.

whereismyjetpack (approved):

PR #1118, opened 2023-01-06
updates redis version

whereismyjetpack:
we're updating redis in production, this is to match in CI and local what we will be running in prod


banukutlu:
We should also update the Gemfile, right?

banukutlu (approved):
👍🏻

banukutlu:
> We should also update the Gemfile, right?

nevermind  this, I just saw your previous PR :) thanks!

PR #1135, opened 2023-01-24
make db connection pool configurable

whereismyjetpack:
```
irb(main):003:0> Rails.application.config_for(:database)
=>
{:adapter=>"mysql2",
 :encoding=>"utf8",
 :host=>"catalog-connpool-mysql-master",
 :pool=>10,
 :username=>"blackcat",
```

we are seeing a rise of connection pool errors in prod, they are happening mostly to bots, but discovered we can't change this on the fly so this PR makes the connection pool configurable. 

we might want to see if we can prevent bot users from interacting with the database. I'm not sure what operations are even happening, we might need to increase logging for db statements to see?

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -26,7 +26,7 @@ production:
   adapter: mysql2
   encoding: utf8
   host: <%= ENV.fetch("MYSQL_HOST", Settings.mysql.host ) %>
-  pool: 5
+  pool: <%= Settings.mysql.connection_pool %>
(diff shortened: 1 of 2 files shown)


banukutlu:
> 
> we might want to see if we can prevent bot users from interacting with the database. I'm not sure what operations are even happening, we might need to increase logging for db statements to see?

that is a good idea, next sprint let's look into this! thanks for bringing it up!

PR #1142, opened 2023-02-13
fix redis deprication

whereismyjetpack:
Fixes redis pipelining in preparation for redis 5. 

We didn't update redis at this time, because it's not compatible with the version of datadog that we are using; and we can't update datadog because of our MonkeyPatching on Datadog.


banukutlu (approved):
👌🏻  thanks!

PR #1144, opened 2023-02-17
Adds alert to settings.yml

ajkiessl:


whereismyjetpack:
@ajkiessl  this can be done in the -config repo -- let me know if you need anything! https://github.com/psu-libraries/catalog-config/blob/main/clusters/prod/manifests/psulib_blacklight/prod.yaml#L42

PR #1156, opened 2023-04-03
Fixes Microform scan requests (and ILLiad on Microforms) not populating

### LINKED ISSUE #1123: Availability: Microform scan requests (and ILLiad on Microforms) not populating OpenURL links
Microform scan requests and ILLiad on Microforms links that substitute in when a Microform is not correctly populating any fields but instead the links just taking patrons to ILLiad.
- Sample record currently with both: https://catalog.libraries.psu.edu/catalog/140905
- Sample with scan only:  https://catalog.libraries.psu.edu/catalog/13855626

- [ ] Fix "Request Scan" links on microforms as an OpenURL which populates at least: title, call no. and has `Genre=GenericRequestMicroScan`

Note: These links were working correctly in the past. https://github.com/psu-libraries/psulib_blacklight/blob/9e95474a993e43571fa10a6acfcbee2ed547f933/app/javascript/availability/components/ill_link.jsx

ajkiessl:
The issue here was the valid_isbn_ssm field was sometimes coming in as an empty array or null value.  I changed the code to better handle this.  I also added a test to catch the issue.

closes #1123

### DIFF
--- a/app/javascript/availability/components/special_request_link.jsx
+++ b/app/javascript/availability/components/special_request_link.jsx
@@ -35,7 +35,7 @@ const SpecialRequestLink = ({ holding, locationText }) => {
           }
         }
       })
-      .catch(() => { })
+      .catch(() => {})
       .finally(() => {
         setShowSpinner(false);
         setUrl(linkUrl);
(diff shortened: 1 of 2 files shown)


banukutlu (approved):
👌🏻  thanks Alex!

PR #1158, opened 2023-04-05
Refactor Fixtures

whereismyjetpack:
Addresses #1159 

- I took all of the json objects in `current_fixtures.json` and created a `spec/fixtures/current_fixtures/#{id}.json`

- Introduced a `bundle exec rake solr:load_from_catalog` rake task that, by default will pull the raw.json from all of the `Settings.fixture_catkeys` on the `Settings.fixture_host` It saves the json to `spec/fixtures/current_fixtures/#{catkey}.json`

- refactored `bundle exec rake solr:load_fixtures` to iterate on all the `.json` files in the `current_fixtures` directory. 

TODO
if we're still generating fixtures from traject, we will need to split them into seperate json files.
Document different ways to get fixture json. the `current_fixtures` directory is the "source of truth" for what our fixtures are. We can add fixtures anyway we see fit, loading from the catalog is just a convenience for when we find records that we want to add to our fixtures (for example, the iif_manifest_url work we've been doing

Lastly, this is all pencil, not pen. Please try this out and give feedback on what does, and does not work.


whereismyjetpack:
Right now, when adding fixtures a lot of our specs will fail because the `numDocs` changes, or the range within call_number_browse changes. I'm not sure the idiomatic way to handle this, do you also change specs when changing fixtures? do you make the specs less specific?

banukutlu:
> call_number_browse

Until now we had to change the numbers on those specs since we were not adding fixtures that often. But as you mentioned making the specs more generic would be more ideal but depending on the spec of course maybe it will still make sense to leave testing the counts 🤔

banukutlu (approved):
🚀  great improvement! thanks Dann!

PR #1168, opened 2023-04-17
"Include all online results" button for campus facet

### LINKED ISSUE #529: Add "add all Online items" to search results already faceted by Library
Users (and the library people helping them) at a particular location can see all items immediately available to them, which means everything which has a location in the library they're currently at and everything that's available Online, since they can immediately access that too. 

In terms of results, this means everything selected by the Library facet, e.g. York, AND everything that is Online. (Maybe some kind of special button up top? "Include all online results"?)

## WDDLL

- [ ] Users who are faceting to a single Library can also add _all_ items Online.

## Example link

Right now there is no way to make the inclusive selection. the ANY choice in Advanced options only applies to the top section. The controlled values are "AND have these attributes" and selecting both Online & Hershey only shows both: https://catalog.libraries.psu.edu/catalog?utf8=%E2%9C%93&op=AND&all_fields=&title=&author=&subject=&identifiers=&series=&publisher=&f_inclusive%5Baccess_facet%5D%5B%5D=Online&f_inclusive%5Blibrary_facet%5D%5B%5D=College+of+Medicine+%28Hershey%29&range%5Bpub_date_itsi%5D%5Bbegin%5D=&range%5Bpub_date_itsi%5D%5Bend%5D=&sort=score+desc%2C+pub_date_itsi+desc%2C+title_sort+asc&search_field=advanced&commit=Search

ajkiessl:
The include button 👇 
Screenshot 2023-04-17 at 4 31 56 PM

And the remove button 👇 
Screenshot 2023-04-17 at 4 33 17 PM

There's no logic to remove the 'add_all_online' query parameter in the URL after clicking the include button when the user leaves the single campus facet search.  However, there is logic to ignore this parameter when the user is not in a single campus facet search.  Therefore, this should not cause any confusion or issues unless the users are very picky about URLs.

closes #529

### DIFF
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -60,4 +60,22 @@ module FacetsHelper
     default_component = facet_config.pivot ? Blacklight::FacetItemPivotComponent : PsulFacetItemComponent
     facet_config.fetch(:item_component, default_component)
   end
+
+  def campus_facet_all_online_links
+    if params['f'].keys == ['campus_facet']
+      if params[:add_all_online].present?
+        link_to(
+          'Remove online results',
+          "?#{params.except(:controller, :action, :add_all_online).to_query}",
+          class: 'btn btn-outline-secondary btn-sm mt-2'
+        )
+      else
+        link_to(
+          'Include all online results',
+          "?#{params.except(:controller, :action, :add_all_online).to_query}&add_all_online=true",
+          class: 'btn btn-outline-primary btn-sm mt-2'
+        )
+      end
+    end
+  end
 end
--- a/app/helpers/facets_helper.rb
+++ b/app/helpers/facets_helper.rb
@@ -62,8 +62,8 @@ module FacetsHelper
   end
 
   def campus_facet_all_online_links
-    if params['f']&.keys == ['campus_facet']
-      if params['add_all_online'].present?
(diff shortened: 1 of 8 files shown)


app/helpers/facets_helper.rb
```diff
@@ -60,4 +60,22 @@ def facet_item_component_class(facet_config)
     default_component = facet_config.pivot ? Blacklight::FacetItemPivotComponent : PsulFacetItemComponent
     facet_config.fetch(:item_component, default_component)
   end
+
+  def campus_facet_all_online_links
```
ajkiessl, line 64: Also, I wasn't too sure if this was the right place to put this method.  It's related to the facets, but isn't actually in the facets drop-down.

ruthtillman:
@ajkiessl I really like that button! How does it look on mobile?

ajkiessl:
@ruthtillman 
Screenshot 2023-04-18 at 10 11 03 AM
Screenshot 2023-04-18 at 10 11 16 AM

app/helpers/facets_helper.rb
banukutlu: I think it makes sense here 👍🏻

banukutlu (approved):
looks great! thanks Alex :shipit:

PR #1172, opened 2023-04-25
Adds javascript to add some aria labels to dropdowns in advanced search

### LINKED ISSUE #304: Chosen-like drop-downs in Advanced search inaccessible
1. "it's a dropdown list but it says it's a button"
1. when tabbing through options inside a dropdown nothing is announced

_from accessibility testing with Michelle on 5/14/19_

Aha! Link: https://penn-state-university-libraries.aha.io/features/STPROJ-41

ajkiessl:
Mac's VoiceOver was able to read all the options in the dropdowns properly.  The top-most and bottom-most dropdowns were being referred to as "buttons".  I was not able to totally get the top dropdown to stop being referred to as a "button".  It is contained within an h3 tag, so the screen reader treats it like a menu dropdown.  I added an aria label to indicate that it's a "listbox", but the screen reader says it's a "listbox, popup, collapsed, button".  Setting the role to listbox here seemed to really confuse the screen reader, so I didn't do that.  I also didn't change the tag from h3 to something else because having it as a header tag forces the screen reader to read the whole element, which is helpful to indicate that there is a dropdown contained within.

The bottom dropdown I was able to fix by setting it's role to "listbox".

closes #304


ajkiessl:
There is no update for blacklight_advanced_search, so that won't help.

ajkiessl:
This is ready for review.

I was not able to fix the keyboard trap by updating the plugins.  I added an issue for this: #1175

Smullz622 (approved):
lgtm

banukutlu:
@ajkiessl could be an obvious answer but I did not have a chance to look into this much, so just to clarify, are the aria labels and role that are being set by the js in this PR, cannot be added by overriding a view?

banukutlu (commented):
@ajkiessl could be an obvious answer but I did not have a chance to look into this much, so just to clarify, are the aria labels and role that are being set by the js in this PR, cannot be added by overriding a view?

ajkiessl:
@banukutlu I potentially could have, but it seemed like the simplest way, and the way with the least amount of complexity and code changes was to use javascript.

I would have had to add several more overrides to do this.  The views being affected in this issue aren't already overriden.

PR #1179, opened 2023-05-05
Adds RIS file export to share dropdown

### LINKED ISSUE #335: Citation functionality
Ruth: I updated our Wiki page about this toward the end of 2022. https://github.com/psu-libraries/psulib_blacklight/wiki/Adding-RIS-Citation-Export

For implementation -- let's replace the Refworks and EndNote exports since they can both use RIS and they're not going to be as good quality as what we're giving in RIS. This will be in the Share dropdown and it'll be just **Email** and **Export RIS**.

Multiple authors can be put in as repeating A2s, however none of the rest of the fields can repeat so we'll need to grab field[0] for any arrays. (did my own test to see how Zotero handles it)

Including below notes from how they did it at Wisconsin -- they're using straight Ruby vs. Blacklight, so we would probably have to look into how BL does it when making exports, but it might be useful. (I got this info when checking something else they do that I decided won't be a good fit for our catalog.) Our own RIS mappings are in the wiki link above.

```
We register the mime type (in the rails default mime_types.rb) with: Mime::Type.register "application/x-research-info-systems", :ris

We have a “show.ris.erb” that looks like:

TY  - <%= @doc.ris_type %>

T1  - <%= @doc.title %>

AU  - <%= @doc.simple_author %>

PY  - <%= @doc.dates.first %>

PB  - <%= @doc.edition_and_imprint %>

AB  - <%= @doc.physical_descriptions.first %>

<% @doc.call_numbers.each do |cn| %>

CN  - <%= cn %>

<% end %>

N1  - <%= @doc.display_notes.join(";") %>

ID  - <%= @doc.id %>

UR  - <%= system_item_url(@doc.id) %>

ER  -

And then link to our ris view of the show page by specifying the format:

<%= link_to t(:cite_export), system_item_path(@doc.id, format: :ris) %>

 ```

Smullz622:
Fixes #335 

With the addition of the RIS file export, we removed the Refworks and Endnote share options. They are no longer supported, and the RIS export makes them redundant.


ajkiessl (commented):
This looks good to me.  The only thing I can think of to consider doing would be to add a feature test to test some of the logic around whether or not the RIS File link shows up in the drop down.  I don't know the details around all the logic there though, so it may not be necessary.  It might be worth testing that the file download is triggered too.

ajkiessl (approved):
🚀

PR #1183, opened 2023-05-09
preview/db searches

whereismyjetpack:
- Revert "Use Redis to store searches (#730)"

Overriding `Search` was causing issues with devise clean users rake task, which we need to run on a frequency to get rid of guest users in the database. 

Since we no longer have the desire to get rid of mysql, this will allow us to run the clear searches, and clear users rake tasks on a schedule to get rid of database bloat. 

this will need to be merged before I can get the workflows for cleaning up users / etc


app/controllers/catalog_controller.rb
```diff
@@ -8,7 +8,7 @@ class CatalogController < ApplicationController
   include Blacklight::Searchable
   include Browse
   include ReportIssue
-  include SearchContext
+  include ::ReportIssue
```
ajkiessl, line 11: Is there a difference between `ReportIssue` and `::ReportIssue`?

ajkiessl (changes requested):
I have one change request regarding the Searches table migration.  Without creating a new migration, running the migrations won't do anything.  Typically it's not a good idea to go back and edit or remove old migrations.

app/controllers/catalog_controller.rb
whereismyjetpack: I really have no idea

whereismyjetpack: looks like this was changed in this PR https://github.com/psu-libraries/psulib_blacklight/pull/855/files I've fixed this to only have the one `include` statement

db/migrate/20210416203208_drop_searches.rb
```diff
@@ -1,5 +0,0 @@
-class DropSearches < ActiveRecord::Migration[6.0]
-  def change
-    drop_table :searches
-  end
-end
```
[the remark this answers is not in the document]
whereismyjetpack: Added a new migration that matches the old one

ajkiessl (approved):

PR #1187, opened 2023-05-10
Refactor IIIF manifest URL checking and tweak error handling behavior

EricDurante:
Here's a light refactoring to get this behavior closer to the data and to improve the test coverage.

I also changed how we handle errors encountered during this process so that problematic URLs are passed on to Mirador, thus allowing it to render error information instead of simply not displaying the viewer at all.

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -39,4 +39,33 @@ class SolrDocument
   # and Blacklight::Document::SemanticFields#to_semantic_values
   # Recommendation: Use field names from Dublin Core
   use_extension(Blacklight::Document::DublinCore)
+
+  def iiif_manifest_url
+    if self[:iiif_manifest_ssim].present?
+      follow_redirects(self[:iiif_manifest_ssim].first)
+    end
+  end
+
+  private
+
+    def follow_redirects(url)
+      host = URI.parse(url).host
+
+      # Skip any further redirect business if we already know that we're going to get a
+      # response with good CORS headers.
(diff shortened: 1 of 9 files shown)


app/models/solr_document.rb
```diff
@@ -39,4 +39,33 @@ class SolrDocument
...
+  end
+
+  private
+
+    def follow_redirects(url)
+      host = URI.parse(url).host
+
+      # Skip any further redirect business if we already know that we're going to get a
+      # response with good CORS headers.
+      return url if ['cdm17287.contentdm.oclc.org', 'digital.libraries.psu.edu'].include?(host)
```
EricDurante, line 56: @whereismyjetpack, I left this domain check in here for now. I think that it's a good idea to add a check for a CORS header in the response also - which I'll do.

My reasoning for checking the domain first is that every web request that we perform here is going to be an opportunity for something that's out of our control to go wrong that could make the whole page slow to load or cause it to explode altogether. I'm assuming that a lot of the manifest URLs in the index are at the Content DM domain and served with permissive CORS headers. So for all of those cases, I figured that it would be better to allow the IIIF viewer to handle any errors on the front end so that _only_ the viewer blows up (instead of the whole page).

Do you think that makes sense?

whereismyjetpack: I think this makes sense. Since we presently only have 6 of these things, it's not clear if most will be content DM etc. I think it looks good the way it is though.

whereismyjetpack (approved):

PR #1189, opened 2023-05-11
Store Bookmarks

### LINKED ISSUE #883: Storing Bookmarks ... or not
Presently, bookmarks are stored in the database, linked to a user accounts. User accounts are only created per session. So as long as the user's session remains active, there is a user account in the database with that session ID, and the associated bookmarks are linked to _that specific session user_. Once the session is lost, when the catalog page is reload or visited again, the application cannot retrieve the original user session record, and the bookmarks are lost.

Technically, neither user nor bookmarks are "lost" because they're still in the database, we just can't retrieve them anymore because we don't have the original session ID that was used to create and link them.

## Current Issues

* the user and bookmark records are never cleaned out, so these records accumulate and are never removed
* bookmarks are only active as long as the user's session is
    * presently, the user's session is stored as a client-side cookie
* it's not clear to the user that bookmarks are ephemeral

## Proposals

Here are some solutions in ascending order of complexity.

### Option 1: Remove Bookmarks

One idea is to remove bookmarks completely. Because the urls for searches and individual records are their own permalinks, the bookmarking feature could be replaced with the bookmarks that your browser keeps.

* helpful UI messaging
* JS integration? i.e., clicking "bookmark" triggers the bookmarking action in your browser

Pros: super easy, simplifies things a lot (no more user records!)
Cons: no bookmarking feature

### Option 2: Alternate bookmarking storage

We can store the bookmarks in Redis, like we do with searches, and give them an appropriate TTL value, ex. 30 days. User records can probably go away as well, since this will all be driven by the session. It's still up to the user to maintain their session in the browser. If that session outlives the Redis TTL, then their bookmarks would disappear. Otherwise, if the user looses their session or starts a new one, the bookmarks will not follow and be "unmoored" in the Redis store to be culled when the TTL expires.

Pros: duplicates searches solution, gets rids of users (?)
Cons: bookmarks still disappear, moderate complexity

### Option 3: OAuth integration

We integrate with OAuth and create user accounts for those that want to login. The user's account is permanent, and bookmarks can be retrieved regardless of session status. Guest accounts would need to be periodically truncated.

Pros: it's got it all
Cons: complicated. We need to manage users (not hard, but just more things to do)

## Related Issues

Implement Blacklight Bookmarks - My Account Permanence #331 
User Management #878

whereismyjetpack:
Requires login for bookmarks feature. Fixes #883


whereismyjetpack:
@EricDurante I asked for your review because i'm doing some funky things that an outsider perspective might be good for. feel free to blindly accept or deny, as well.

whereismyjetpack:
They get stored in the database. presently to a "guest" user that is bound to the session. This change will bind them to the proper user record, and be bound to the user.

ajkiessl (approved):
This LGTM.  I don't see the "Login to Bookmark" link in the preview instance when not logged in.  Idk if it's still pulling my user from the header but I'm not fully authenticated.  Idk if this is a big deal, though.  The "Bookmarks" link is still there and it authenticates me when I click it.

ajkiessl:
Ah nevermind I'm looking in the wrong place.  It does show up on the items.

EricDurante:
Hey @whereismyjetpack, sorry that I didn't get a chance to take a closer look at this. I've been tied up quite a bit the past few days. Were there any specific decisions that you wanted another brain on?

whereismyjetpack:
This is ok! was a bit out of my comfort zone with passing the fullpath parameter, and redirecting. I couldn't find a better way to do it, and it's tested to work -- but i'm not sure if there is a more idiomatic way to do that portion.

EricDurante:
@whereismyjetpack 
Oh, gotcha. I guess probably the way that I'd normally do this would be to have the "Login to Bookmark" button not be a button to go create a new user session, but rather to just go and do the bookmarking. So the button would instead be "Bookmark this here thing" regardless of whether the user was already signed in or not, and it would be handled by the BookmarksController. The action in the BookmarksController would check to see if the user was singed in. If they were, it would create the bookmark. Otherwise it would store the location (in the session) and redirect to the login action. Then the login action would redirect back to the location stored in the session.

Maybe there's a reason why that approach wouldn't work (or make sense) here - I'm not sure. Either way, I don't think there's anything wrong with the way that you did it. The only thing that I'd suggest changing maybe would be to use view helpers to construct the login link - and particularly to construct the link's href.

Does that help, or do I seem out of my element?

PR #1199, opened 2023-05-18
Thesis Department display

### LINKED ISSUE #765: Add linked search for thesis department or departments
This feature allows someone to view all theses from just a particular department. It's a helpful way to get an overview of peer work or share departmental outcomes with a saved search, etc. Use field indexed in https://github.com/psu-libraries/psulib_traject/issues/305

Display in top area of record directly under Additional Creators because, if it exists, it's an important data point. It will only exist for these kinds of records.

Link text to _facet search (just like linked subject or creator search).

We may look later into adding it to the Advanced Search page or something like that. because the data probably varies widely it's not going to be great, but it should probably be a controlled drop-down driven by the data and that could be a lot. I'll want to analyze etc before that.

Requires https://github.com/psu-libraries/psulib_traject/issues/305

ajkiessl:
Display thesis departments as a list of links to a facet search for the thesis department.

Also, I refactored a number of link generating methods into their own helper module.  They were sharing a lot of code, so I put most of the logic for generating the links into a single function.

closes #765 

traject side: https://github.com/psu-libraries/psulib_traject/pull/473

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -205,6 +205,7 @@ class CatalogController < ApplicationController
     config.add_facet_field 'subject_browse_facet', show: false, limit: 0
     config.add_facet_field 'subject_facet', show: false
     config.add_facet_field 'title_sort', label: 'Title', show: false
+    config.add_facet_field 'thesis_dept_facet', show: false
 
     #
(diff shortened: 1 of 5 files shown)


Smullz622 (approved):
lgtm

PR #1207, opened 2023-05-24
Don’t show links to manifests if we’re going to render an IIIF viewer

EricDurante:
...as requested by @ruthtillman a couple of weeks ago.


ajkiessl (approved):
LGTM

PR #1208, opened 2023-05-25
Add support for records with multiple IIIF manifests

EricDurante:
This should work pretty well for any future records that have up to a handful of different IIIF manifests. If any records ever have more than that, then we'd probably want to do something a bit different with the UI.

@whereismyjetpack It would be nice to be able to preview this somehow so that everyone can see what it will do and give any feedback...


ajkiessl:
@EricDurante If you rename the branch to something starting with `preview/`, it should deploy a preveiw after CI passes.  So maybe something like `preview/multi-iiif` ?

EricDurante:
Thanks, Alex. It looks like Dann already did that - thanks, Dann!

I'm not sure what/how data is used in the preview environments, but there's only the one fabricated example of a multiple-manifest record in the solr fixtures at the moment (/catalog/7280921), and it only has two manifests.

whereismyjetpack:
preview envs by default will load fixtures, so it is there https://catalog-multiple-iiif.dev.k8s.libraries.psu.edu/catalog/7280921

ajkiessl (approved):
LGTM

PR #1221, opened 2023-07-19
updating browser tools orb to successfully download chrome drivers

bryanlopezrr:
The current version of orb fails to install the correct chrome driver. The latest release has a solution. You can find the details here: https://github.com/CircleCI-Public/browser-tools-orb/releases

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,9 +77,6 @@ jobs:
     steps:
     - browser-tools/install-chrome
     - browser-tools/install-chromedriver
-    - run:
-        name: Remove Cruft
-        command: rm *
     - checkout
     - restore_cache:
         keys:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   docker: circleci/docker@1.5.0
-  browser-tools: circleci/browser-tools@1.4.0
+  browser-tools: circleci/browser-tools@1.4.3
```
whereismyjetpack, line 5: It looks like the Remove Cruft stage is now redundant and is failing the build.

whereismyjetpack (changes requested):

whereismyjetpack (commented):
CI looks good, but the tests are still failing because the webdrivers gem is trying to download a newer version. the webdrivers gem has yet to be updated. I fear for now we will have to version pin both webdrivers gem, and circleCI configuration to 114.x

PR #1226, opened 2023-09-11
updates browser-tools orb

whereismyjetpack:


whereismyjetpack:
@ajkiessl - I believe the failing spec to be a flaky one. I'm not sure if we should try to fix it, or xit.. what do you think?

ajkiessl:
I pushed a change that should fix this and still test (I'm assuming) the same thing.  The `visually-hidden` class for that element was changed to `sr-only` at some point.  I think these accomplish the same thing.

spec/fixtures/vcr_cassettes/viewing_a_record/when_the_record_has_an_IIIF_manifest_URL/renders_the_Mirador_viewer_with_no_workspace_type.yml
```diff
@@ -94,4 +94,106 @@ http_interactions:
       encoding: ASCII-8BIT
       string: var _GBSBookInfo = {};
   recorded_at: Thu, 25 May 2023 18:31:49 GMT
+- request:
+    method: get
```
whereismyjetpack: I wonder if we should do what we do In scholarsphere for this? 

https://github.com/psu-libraries/scholarsphere/blob/main/bin/ci-rspec#L54

I'm not sure we should be recording calls to get chromedriver, we also may want to allow chromedriver.storage.googleapis.com in our allowed hosts configuration?

whereismyjetpack: Oh i see you did revert this change.

ajkiessl (approved):

PR #1238, opened 2023-10-27
Shakapacker migration

### LINKED ISSUE #1236: psulib_blacklight security vulnerabilities

Smullz622:
Fixes #1236 

migrating from web packer to shakapacker

Notes: getting jQuery to work was the biggest hurdle. Eventual solution was to move the logic that adds jQuery from `config/webpack/environment.js` to `config/webpack/webpack.config.js`


whereismyjetpack:
I'd feel better merging this if we had a preview to test first. I'll fork this branch and deploy so we can test that the asset compilation works in k8s

Smullz622:
Sounds good, thanks @whereismyjetpack

whereismyjetpack:
This looks good to me, https://catalog-shak.dev.k8s.libraries.psu.edu/

whereismyjetpack (dismissed):

ajkiessl (commented):
This is a bit picky, but there's a reference to webpacker in the README: "Because we use webpacker, when upgrading the Blacklight gem for this application, we should also update package.json to the same version as the installed gem so as to keep these things in step from Blacklight."

Otherwise, this looks good.

ajkiessl (approved):
:shipit:

PR #1251, opened 2023-12-01
ILL placement for closed libraries

### LINKED ISSUE #1211: Research: Sending Closed Libraries to ILL Placement?
Right now, we have 2 libraries that are closed. I got a request for whether we could expand the ILL routing for materials that are only at or only on-shelf at one of these closed libraries. Because of our floating collection, we can't actually use Locations for this -- plus it's so big that doing it at the Library level makes more sense. Closed libraries are FAYETTE and WSCRANTON. (updated by Chris)

How feasible would it be to include this in our logic? If it's holdable: true, but the only available copies are:

* at Library in Closed Library List OR
* at Location in Non Holdable Locations list

then route it to ILL. Chris H is good with this.

Smullz622:
Fixes #1211 

Determined at the library rather than location level, holdable items that can be found in currently closed libraries but are not in another holdable location are directed to ILL placement.

### DIFF
--- a/app/javascript/availability/index.js
+++ b/app/javascript/availability/index.js
@@ -21,6 +21,7 @@ const availability = {
   reserveCirculationRules: reserveCirculationRules.reserve_circulation_rules,
   movedLocations: [],
   nonHoldableLocations: locations.non_holdable,
+  closedLibraries: locations.closed_libraries,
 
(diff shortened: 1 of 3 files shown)


ajkiessl (approved):
LGTM!

PR #1252, opened 2023-12-05
allow call number display for ILLEND items

### LINKED ISSUE #1246: Remove logic suppressing Call # display for ILLEND items
We initially weren't showing items checked out to ILL Lending call numbers... I don't remember exactly what went into it, I think it might've been because it wasn't in the original CAT. Based on feedback I've gotten, this is worrying some folks for no good reason (sending librarians off to make sure there IS a call number). Removing it will also reduce our exceptions by one, which is always nice.

Can we remove the exclude so that call number displays as normal? The logic itself appears to be in here:

https://github.com/psu-libraries/psulib_blacklight/blob/main/app/javascript/availability/components/holding_details.jsx#L7-L10

Sample catkeys:

* 39786
* 177533
* 370199
* 427617
* 617500

Smullz622:
Fixes #1246 

Suppressing call number for ILLEND items is no longer needed so that logic was removed. Also updated a flaky spec that relied on the number of fixtures never changing.


spec/features/availability_spec.rb
```diff
@@ -151,8 +151,9 @@
 
   describe 'User visits catalog record page', js: true do
     it 'that has holdings to display' do
-      visit '/catalog/1839879'
-      expect(page).to have_css 'div[class="availability"][data-keys="1839879"]'
+      visit '/catalog/370199'
```
ajkiessl, line 154: I'm curious why the record we are testing here needed to change?

Smullz622: Technically, it probably didn't. I just felt better using a record with ILLEND as the current location at least for initial testing to be sure that is displaying now, but it's subject to change once that item is returned so I guess it doesn't really matter.

ajkiessl: Ah okay 👍

ajkiessl (dismissed):
Looks good!  Just one suggestion, but it doesn't need to be changed.

spec/features/prev_next_spec.rb
```diff
@@ -9,12 +9,14 @@
       fill_in 'q', with: ''
       click_button 'search'
       click_on 'An American marriage / a novel by Tayari Jones'
-      expect(page).to have_content '← Previous | 4 of 597 | Next →'
+      expect(page).to have_content '← Previous | '
+      expect(page).to have_content ' | Next →'
```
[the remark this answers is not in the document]
Smullz622: Oh yeah, that's a better idea. Done!

PR #1254, opened 2023-12-08
Remove stopwords from title browse

ajkiessl:
Fixes at least part of https://github.com/psu-libraries/psulib_traject/issues/491

Strip "An", "A", and "The" from the beginning of title browse searches.  I'm not going to close https://github.com/psu-libraries/psulib_traject/issues/491 since there may still be more we can do.  I'll add a comment to it, though, to indicate what has been done.

One thing to note about this: If someone types "An American Marriage" in the search field and hits enter to trigger the search, the value in that field will change to "American Marriage".  The stripping of "An" will be shown to the user.  Not sure if this will be a good or bad thing.  On one hand, it will indicate to the user that the word has been stripped from the search.  On the other hand, some users might be taken aback by the search forcefully removing words for them.


app/controllers/browse_controller.rb
```diff
@@ -44,6 +44,8 @@ def subject_list_params
     end
 
     def title_list_params
+      params[:prefix] = params[:prefix].gsub(/^(?i)(an|a|the) /, '')
```
Smullz622, line 47: I think this needs to be `&.gsub`. When I go to `All` in the browse, it throws a nil error.

Smullz622 (changes requested):
One quick thing but looks good overall

app/controllers/browse_controller.rb
ajkiessl: Ah right, I'll add that.

Smullz622 (approved):
lgtm!

PR #1255, opened 2023-12-11
serials edge case fix

Smullz622:
fixes https://github.com/psu-libraries/psulib_traject/issues/492

Previously summaries, indexes, and supplements info was only being pulled for the first holdings statement for a given library/location pair. 

When there is only one holdings statement, the display is unchanged. When there are multiple holdings statements for a library/location pair, the call number is displayed for each holding statement with summaries, indexes, and supplements nested in a bulleted list beneath.


ajkiessl (approved):
This looks good to me.  The only thing I'd point out is that it looks like the style: 

{
                      listStyleType: 'none',
                      padding: 0,
                      marginLeft: '0',
                    }

Is reused a few times.  It could probably be refactored into a constant.

ajkiessl (approved):
LGTM

PR #1256, opened 2023-12-13
refactored bulk ris export

Smullz622:
This is in prep for the checkouts RIS export in myaccount https://github.com/psu-libraries/myaccount/issues/675. MyAccount can get item ids of checkouts, but ids will have to be passed to blacklight for the full solr metadata needed for RIS

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -39,7 +39,7 @@ Rails.application.routes.draw do
     collection do
       get 'initialize_bookmark'
       delete 'clear'
-      match 'bulk_ris', to: 'bookmarks#bulk_ris', via: [:get, :post], as: 'bulk_ris_solr_document'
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -39,7 +39,7 @@
     collection do
       get 'initialize_bookmark'
       delete 'clear'
-      match 'bulk_ris', to: 'bookmarks#bulk_ris', via: [:get, :post], as: 'bulk_ris_solr_document'
+      match 'bulk_ris/:item_ids', to: 'bookmarks#bulk_ris', via: [:get, :post], as: 'bulk_ris'
```
ajkiessl, line 42: I don't think `:item_ids` is needed here since we aren't using params in the path. I may not fully understand how this works though.

ajkiessl: Ah, sorry.  I understand now: myaccount does that.

ajkiessl (approved):
Looks good!

PR #1259, opened 2024-01-02
Bookmark All On Page

### LINKED ISSUE #1212: Explore: is it possible to select a whole page to bookmark?
This is a question we had, it's not high priority but it also makes sense when supporting lit review searches to let people save a bunch of promising results to bookmarks and then weed them out.

Is it feasible? Would it add too much complexity? Is it the kind of thing where we could check all and let someone uncheck before adding to bookmarks or would _that_ add too much complexity.

ajkiessl:
closes #1212 

I few things to note:

- I modeled the javascript code for submitting all the Bookmark checkboxes on the page after this code in the base blacklight 8 code: https://github.com/projectblacklight/blacklight/blob/main/app/javascript/blacklight/checkbox_submit.js
- There's so much going on in this code that I decided it would be best to just do a single feature test with rspec that goes through all the functionality for the "Bookmark All On Page" button.  This tests ends up testing the orginal bookmarking functionality, so I removed the old test.
- The linter was trying to get me to remove things that would break the code.  I had to update the [[.eslintrc.json](https://github.com/psu-libraries/psulib_blacklight/pull/1259/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090f)](https://github.com/psu-libraries/psulib_blacklight/pull/1259/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR20) file to use a new version of "ecma" to get it to work.  Editing this file caused VSCode to format the JSON in it ( I just kept that).


Smullz622 (approved):
LGTM!

PR #1270, opened 2024-01-10
refactor specs

Smullz622:
Changing complier strategy from digest to mtime to improve speed. Brings test suite time down from ~28 mins to 8 mins.


ajkiessl (approved):
Great! It would be interesting to see if this helps anywhere else (like ETDA).  Glad it was just a one line fix.

PR #1294, opened 2024-04-15
Convert HathiLinks to API Calls

### LINKED ISSUE #1291: Change HathiTrust links to API Call
## Background

The earliest Catalog used the HathiTrust API. It's been 5 years so we can't just roll it back, but it might be worth looking at.

During lockdown, HathiTrust provided us with a CSV datafile matching our catalog holdings and their scans. We had indexed this along with our MARC data to provide keys which saved us the trouble of API lookups. At the time, we were using ETAS (Emergency Temporary Access Service) to access and digitally check out scans of books that are in copyright but were locked in our closed library. When they stopped providing ETAS checkouts, we switched to only show items which are in the public domain.

HathiTrust stopped sending us matching CSVs in July 2021. Since we were mostly concerned with pre-copyright books, this hasn't been a thing that updates much, but more books have come into public domain since then, more have been identified as out of copyright by teams working for HT, and some records have changed over time.

## Current Behavior

We currently use data from our index to determine whether or not we can link to a HathiTrust scan of the book. 

## New Behavior

We'll want to send an API request similar to what we do with Google Books. We **only** want to display the links where we get full text copies. HathiTrust's search-inside-only functionality is far inferior to the Google Books search inside and we don't ever want to include it.

For the sake of being sure that updating this doesn't mean that we accidentally mess up interacting logic with Google Books, display logic should be:

- If an item is ONLINE then we don't display HathiTrust or Google Books links.
- If we get a full copy from HathiTrust, generate link and don't link to Google Books
- If we don't get a full copy from HathiTrust, then IF there's a full or preview copy from Google Books (current query is fine) we generate that link.
- If we don't have either a full HathiTrust copy or a full or preview copy from Google books, don't display any special links.

### HathiTrust API

The API uses only one identifier parameter. Choose the first one which exists in the following order:

- oclc
- lccn
- issn (first)

Syntax is: `https://catalog.hathitrust.org/api/volumes/brief/oclc/424023.json` with appropriate param in place of oclc if needed

We are looking for at least ONE item which has: `"usRightsString":"Full view"`

Sample with several items in Limited View only: https://catalog.hathitrust.org/api/volumes/brief/oclc/424023.json

Sample with only one item, Full View: https://catalog.hathitrust.org/api/volumes/brief/oclc/433.json

These two pages have more documentation:

- https://www.hathitrust.org/member-libraries/resources-for-librarians/improve-discovery/integrate-records/
- https://www.hathitrust.org/member-libraries/resources-for-librarians/data-resources/bibliographic-api/

## Not Changing

We are not changing our indexing/indexed data at this point.

ajkiessl:
fixes #1291


ajkiessl:
preview here: #1297

PR #1295, opened 2024-04-19
WIP upgrade to blacklight 8

Smullz622:


config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
```
ajkiessl, line 1: This file can be removed since we're on rails 7 now.

ajkiessl (changes requested):
So far this looks good.  There's a lot here so I'm kind of skimming over a lot of things, but there was only one thing that stood out to me.

Smullz622:
Finally ready for review! One thing to note, homepage doesn't have facets (Ruth is good with this change) but it's just blank space where they should be for now. I'm not going to focus on fixing that yet because it might get resolved in the bump to blacklight 9 I'm working on next. If that doesn't resolve it or we don't get that complete before the end of this cycle, I'll create a separate priority ticket to resolve that spacing before deploying.

app/views/catalog/_document_list.html.erb
```diff
@@ -0,0 +1,13 @@
+<!--
+# <% view_config = local_assigns[:view_config] || blacklight_config&.view_config(document_index_view_type) %>
+# <div id="documents" class="documents-<%= view_config&.key || document_index_view_type %>">
+#   <%= render documents, as: :document, view_config: view_config %>
+# </div>
+-->
```
jlandiseigsti: Can we lose this commented section?

spec/views/shared/_header_navbar.html.erb_spec.rb
```diff
@@ -8,7 +8,8 @@
   before do
     stub_template 'shared/_user_util_links.html.erb' => 'Some content'
     stub_template 'catalog/_search_form.html.erb' => 'Other content'
-    assign(:search_bar, Blacklight::SearchBarPresenter.new(controller, blacklight_config))
+    # "The SearchBarPresenter has been deprecated. Call `render Blacklight::SearchBarComponent.new' instead"
+    # assign(:search_bar, Blacklight::SearchBarPresenter.new(controller, blacklight_config))
```
jlandiseigsti: Can this be removed?

spec/controllers/catalog_controller_spec.rb
```diff
@@ -4,12 +4,6 @@
 
 RSpec.describe CatalogController do
   describe 'index action' do
-    it 'gets the homepage and renders only the homepage facets' do
-      get :index
-      configured_home_page_facets = %w[access_facet format campus_facet media_type_facet classification_pivot_field]
-      expect(assigns(:blacklight_config)[:facet_fields].keys).to eq(configured_home_page_facets)
-    end
-
```
jlandiseigsti: Will we want this test back when we upgrade to Blacklight 9? (Or am I thinking of the wrong facets?) If so, should we comment this out with a note?

jlandiseigsti (changes requested):
I only found minor things!

spec/controllers/catalog_controller_spec.rb
Smullz622: TBD whether we will want it back. I removed this when we thought they were definitely going away. Not sure yet if they will be coming back with 9

jlandiseigsti (approved):
This looks good to me!

ajkiessl (approved):

PR #1301, opened 2024-05-02
Update HENISCH location

ruthtillman:
We had a request to update HENISCH to "Temporarily Unavailable."


ajkiessl (approved):

Smullz622:
@ruthtillman if it's temporarily unavailable, do we need to add it to the closed libraries list as well?

ruthtillman:
@Smullz622 it's already a special collections location so it's non-holdable etc. It may still get the Aeon link generation, but I don't think we have/should have a sub-list of special collections locations without Aeon links. The description should be enough that someone can point back to it if people request it and say "sorry, as we said, that's not available"

PR #1303, opened 2024-05-06
1053 limit search terms

### LINKED ISSUE #1053: Limit search terms
If a guest types a large search query, our solr instances will OOM, I have a message out to Radu asking about what solr config mitigations we can put in place, but perhaps we should also limit the number of terms a guest can enter into our search box? 

@ruthtillman  -- I recall you mentioning a limit another school was setting? I could be remembering wrong.

jlandiseigsti:
closes #1053 

For edismax searches, we now truncate the number of search terms that we send to solr. This number is dependent on search field...for example, Title may still include many, whereas "All Fields" is capped at 10. This is to prevent solr failures caused by large numbers of search terms. The values for the fields are defined on the ticket.

### DIFF
--- a/app/models/clause_count_limiter.rb
+++ b/app/models/clause_count_limiter.rb
@@ -1,23 +1,21 @@
+# frozen_string_literal: true
+
 module ClauseCountLimiter
   # Limits the number of clauses allowed in a search field. Varies by search field type.
   def limit_clause_count(solr_parameters)
-    return unless is_query_present?(solr_parameters)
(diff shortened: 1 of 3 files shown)


Smullz622 (commented):
Looks great overall! Might be a good idea to add a feature test as well to make sure this is working end to end, though maybe just testing one case, like 'all fields', since the model spec covers logic with different cases

Smullz622 (approved):

whereismyjetpack (approved):

PR #1311, opened 2024-05-20
switch to read only 

Smullz622:
for symphony saas migration, hides I want it button

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -26,6 +26,7 @@ solr:
   # password: 
   # url: http://127.0.0.1:8983
   collection: psul_catalog
+readonly: true
 
 # Used with rake solr:load_from_catalog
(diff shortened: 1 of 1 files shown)


whereismyjetpack:
This can be done dynamically through `config` by setting the `SETTINGS__readonly` environment variable to true and restarting the deployment. We may want to do this instead of cutting a release for each config change.

Smullz622:
@whereismyjetpack that would be done [here](https://github.com/psu-libraries/catalog-config/blob/d8bc9ac76e53d044fa4d0f0e6b23227212ce8716/clusters/prod/manifests/psulib_blacklight/prod.yaml#L55) correct, similar to `hide_annoucement`

whereismyjetpack:
Yup. you could do it there, too!

Smullz622:
doing this in config instead

PR #1321, opened 2024-06-24
upgrade blacklight to latest version of 7

Smullz622:
resolves a lot of security issues/depreciations and prepares for upgrade to BL 8 (on hold for now)


ajkiessl (changes requested):
Same review as the other branch.  Just remove the `config/initializers/new_framework_defaults_6_1.rb` file.  Looks good in the preview too.  If Ruth/Kevin think it's good, then I'd say it's good to ship after removing that file

Smullz622:
> Same review as the other branch. Just remove the `config/initializers/new_framework_defaults_6_1.rb` file. Looks good in the preview too. If Ruth/Kevin think it's good, then I'd say it's good to ship after removing that file

There's one of those for 6.0 and 5.2 so I'll remove those as well. Kevin gave the thumbs up this morning.

ajkiessl (approved):

PR #1329, opened 2024-07-11
make leisure reading room non-holdable

Smullz622:
Leisure reading room will temporarily close for renovation on 7/17

To be deployed 7/17/24


jlandiseigsti (approved):
Looks good!

PR #1331, opened 2024-07-19
adds readability widget

Smullz622:
Adds the [readability widget](https://github.com/NCSU-Libraries-UX/readability-widget) developed by NC State Libraries

The only customization at this time is changing to psu brand colors. 

Current fields are warm background, highlight links, open dyslexic font, and remove images. There's potential in the future to customize further by changing the fields you can toggle on/off (ex. ADHD friendly fonts, etc)


ajkiessl (changes requested):
I think we can remove widget.css and widget.css.map.  Ultimately, shakapacker should compile the widget.scss into the application.css and application.css.map and throw that in the public/packs/css directory.  Otherwise this looks good!

ajkiessl:
Also, this should probably be removed in widget.scss:
```
/*
* to compile run this code from the root of this app
* sass --watch widget.scss:widget.css
*
```
Since shakapacker will handle this.

ajkiessl (approved):
Looks good!

ruthtillman:
Notes on review!

1. When we use "Hide images" there isn't any alt text display -- thinking the Libraries in the header, the cover if there is one, and the footer image. I don't know if we want this per se but the links are still there and functioning. I also don't know from an accessibility perspective if this is at all useful. But it's a thing I noticed. If we want to offer Libraries branding at the top in particular, that's getting lost.
2. The OpenDyslexia font makes the placeholder image for items without covers go squirrely. Is there any way we could exempt it?
open_dyslexic_font_no_google_cover

Smullz622:
There was a user request to have a dark mode option we could consider adding here

ruthtillman:
@Smullz622 I'll leave that to your assessment on feasibility -- I'm sure it'd be popular if we released it.

PR #1332, opened 2024-07-25
turns on crawler_detector

whereismyjetpack:
blacklight will not store searches in the searches table if the crawler_detector evaluates to true. this will prevent writing needless amounts of data to the database. we may want to tweak the lambda to include AI bots as well. I'm going to run a report tomorrow.


Smullz622 (approved):
LGTM!

PR #1338, opened 2024-08-28
Removed WSCRANTON from 'closed_libraries' 

ajkiessl:
Scranton is back online.  I believe this should add Scranton back to allow holds.

Note: The tests I changed technically weren't broken.  I just did a text search for 'WSCRANTON' and figured I'd change those to be consistent.

### DIFF
--- a/app/javascript/availability/libraries_locations.json
+++ b/app/javascript/availability/libraries_locations.json
@@ -802,7 +802,6 @@
   },
   "closed_libraries": [
     "FAYETTE",
-    "WSCRANTON",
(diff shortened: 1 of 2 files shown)


ajkiessl:
We'll see if this works.  There are still a lot of what I believe to be subdivisions of Scranton in our `non_holdable` list (ie "INDEXES-WS", "MEDIA-WS", "REFDESK-WS", and more).  These might need removed from that list too, but idk.

jlandiseigsti (approved):
👍

PR #1340, opened 2024-08-28
Remove Scranton locations from non-holdables

ajkiessl:
I received confirmation to only remove GRFCNOV-WS from the non-holdable list.


jlandiseigsti (approved):
👍

PR #1367, opened 2024-12-18
upgrade to ruby 3.1.6

Smullz622:


.circleci/config.yml
```diff
@@ -76,6 +77,11 @@ jobs:
       SETTINGS__solr__username: catalog
       SETTINGS__solr__password: catalog
     steps:
+    # - run:
```
jlandiseigsti, line 80: Do you think we'll turn uncomment this at any point? If not, we could just remove it.

jlandiseigsti (dismissed):
👍

.circleci/config.yml
ajkiessl: I think we should remove this.

ajkiessl (changes requested):

.circleci/config.yml
Smullz622: Good catch guys. That was leftover from some troubleshooting

ajkiessl (approved):

PR #1389, opened 2025-03-03
moved next button and fixed gap tags

### LINKED ISSUE #1376: Navigation behaving not quite as expected
We had a question from a librarian about why our page navigation looks like:

Image

## Pagination

Agreed this is not intuitive and the ... feels like it should be clickable.

Ideally, the final page would be on the end of the list as a "jump to last."  Similarly, the 1 would remain up front. We could retain the ... as a helpful separator.

So:

1 2 3 4 5 ... 128

or 1 ... 7 8 9 10 11 ... 128

## Previous / Next Navigation

Similarly, we the Previous/Next navigation is on the left side. How hard is it to split to put each on the side of the "direction" you'd expect to move? This isn't as important as pagination but would be nice to have.

granta03:
Fixes #1376. Also reduced window size from 4 to 3 per Ruth's request.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -29,7 +29,7 @@ ActiveRecord::Schema[7.1].define(version: 2024_04_01_150060) do
     t.string "service_name", null: false
     t.bigint "byte_size", null: false
     t.string "checksum"
-    t.datetime "created_at", precision: nil, null: false
+    t.datetime "created_at", null: false
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
   end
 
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -29,7 +29,7 @@
     t.string "service_name", null: false
     t.bigint "byte_size", null: false
     t.string "checksum"
-    t.datetime "created_at", null: false
+    t.datetime "created_at", precision: nil, null: false
```
Smullz622, line 32: this probably got added when tests ran, let's remove it

Smullz622 (changes requested):
One small change but otherwise looks great!

Smullz622 (approved):
lgtm, thanks!

PR #1398, opened 2025-03-07
Add locations to microform listings

### LINKED ISSUE #1388: Availability: change microform scan link display to improve patron expectations
Met with Ashoo & Eric to discuss a request for a change to our display of Microform scanning links. Right now, the item location is overridden by our ILLiad scan request logic and does not display the current location. I think this may date back to 2020 when we were prioritizing scanning service. This change would allow patrons to see that they are requesting a scan of a checked out item and set expectations a little better.

Nothing about the ILLiad link itself would change, it would just be a two-line currentLocation `<br />` link display that mirrors the map scan link here: https://catalog.libraries.psu.edu/catalog/1398320

Mockup:

Image

memsdm05:
Fixes #1388

### DIFF
--- a/app/javascript/availability/components/location_info.jsx
+++ b/app/javascript/availability/components/location_info.jsx
@@ -8,9 +8,8 @@ const LocationInfo = ({ holding }) => {
       ? availability.allLocations[locationID]
       : '';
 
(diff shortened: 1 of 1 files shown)


Smullz622 (approved):
LGTM thanks!

PR #1399, opened 2025-03-07
Add catkey to ILLiad request

### LINKED ISSUE #1377: Add catkey to ILL request
We had a request from our ILLiad coordinator to add catkey to the Micro and Maps record requests. Details of the request are below.

>The links we have going to ILLiad from the catalog Micro and Maps records, would it be possible to pass the catkey as well in the URL? I have set ILLiad up to accept "&catkey=________" This will allow a link to be included in the request sent to those departments that links to the catalog record that the user was on when requesting the material.

>This would be for those requests that use OpenURL links from the catalog record for scans of maps or micro, not the "I Want It" button integration.  We don't need this info for the API / I Want It button requests on checked out items.

memsdm05:
Fixes #1377


ajkiessl (approved):
LGTM

memsdm05:
Going to hold off merge to fix failing check.

PR #1400, opened 2025-03-11
changed caption to invisible header

### LINKED ISSUE #1392: correct placement of alt text header in View Availability table
In the View Availability table, there's an alt text header 'where to find this item in the library', but it's not showing up at the top like a proper header. Instead it reads in the middle of the table after the column headers.

granta03:
fixes #1392


ajkiessl (approved):
lgtm

PR #1401, opened 2025-03-12
added alt text to previous and next buttons

### LINKED ISSUE #1393: improve 'previous' and 'next' alt text
The 'previous' and 'next' links at the top of the search results page would make more sense for a screen reader if the alt text read 'previous result' and 'next result', especially on the individual result page.

granta03:
Fixes #1393 . If the current page is the first or last page, the alt text isn't changed and says "&larr Previous |" and "| Next &rarr", since it only affects every other page due to links. Every other page has the alt text saying previous result or next result as intended. Figured I should point that out.

### DIFF
--- a/app/components/blacklight/search_context_component.html.erb
+++ b/app/components/blacklight/search_context_component.html.erb
@@ -1,0 +1,10 @@
+<div class='pagination-search-widgets'>
+
+  <div class="page-links">
+    <%= link_to_previous_document(aria: { label: 'Previous result' }) %> |
+
+    <%= item_page_entry_info %> |
+
+    <%= link_to_next_document(aria: { label: 'Next result' }) %>
+  </div>
+</div>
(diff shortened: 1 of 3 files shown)


ajkiessl:
I'm not seeing the previous or next buttons anymore on the show page:

Screenshot 2025-03-13 at 10 58 42 AM

I think it should look like this:

Screenshot 2025-03-13 at 11 00 38 AM
Something might be going wrong with the override.  Is it like this for you locally?

granta03:
I think I know why. It was working fine in my first commit but I didn't check if my linter changes messed things up. I'll look at it Monday.

granta03:
Talked with Alex earlier and mine is showing correctly. Not sure what was going on there. The pagination in the individual result still needs fixing so I'm working on that.

ajkiessl (approved):
LGTM

PR #1408, opened 2025-03-21
added header to access action buttons easier

### LINKED ISSUE #1395: consider how to make share, bookmark, and report an issue buttons more obvious for screen readers
The share, bookmark, and report an issue button on an individual record page are easy to miss with the screen reader. Typical screen reader use would be to jump to the record title and then tab through, skipping over those buttons entirely. Michelle's suggestion was to move them lower to be below the initial record info. I'm hesitant to do that from a general UI perspective so my instinct is to leave it as is. To be clear, the screen reader can still reach those buttons, you just have to tab through to get to them.

Ruth's feedback:
I agree with you that I hesitate to move it from a UX perspective, especially with everything else going on. It feels like some of the challenge is something we face with every user which is "how much are they going to skim the site and how much are they going to pay attention to what they could actually do." e.g. if/once a screen reader user notices we have Bookmark, they should be able to navigate to it and we'll add the better alt text so they can be sure it applies to the right item. Which feels not unlike pointing out something on a screen that a visual user never knew we had even though they've used it a hundred times because they're always skimming.

granta03:
Fixes #1395

### DIFF
--- a/app/views/catalog/_record_toolbar.html.erb
+++ b/app/views/catalog/_record_toolbar.html.erb
@@ -1,4 +1,5 @@
 <div class="col record-actions ">
+  <h2 class="sr-only">Actions for <%= document_presenter(@document).html_title.split('/').first %></h2>
   <div class="float-right">
     <div class="d-inline-block mb-1" role="group" aria-label="Actions including with nested dropdown">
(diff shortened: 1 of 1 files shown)


granta03:
fetched new changes from main, used in a new branch

PR #1410, opened 2025-03-24
added aria label to view availability btn w/ title

### LINKED ISSUE #1391: add title to alt text for 'View Availability' and 'Bookmark' in search results
In search results, there is a 'View Availability' and 'Bookmark' button for each result. It's more obvious which record they refer to when tabbing through, but when links on the page are listed, they all just say the same thing. We want the alt text for these buttons to include the title they are associated with.

granta03:
fixes #1391

### DIFF
--- a/app/views/catalog/_index_availability.html.erb
+++ b/app/views/catalog/_index_availability.html.erb
@@ -4,7 +4,9 @@
   <div class="row">
     <div class="availability-button col-md-auto pr-0 invisible">
       <p>
-        <button class="btn btn-primary header" type="button" data-toggle="collapse" data-target="#availability-<%= id %>" aria-expanded="false" aria-controls="<%= id %>-location">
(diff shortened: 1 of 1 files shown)


granta03:
Still trying to figure out how to change the ARIA label based on the current state of the bookmark with javascript. The view availability aria label is good, and I have a placeholder ARIA label for the bookmarks button. As of now, in VoiceOver, when you highlight the button, it first says the ARIA label "Bookmark button for title, group". If you move through the group, it will then say the current state, either "In bookmarks" or "Bookmark". I'll still look into making the label dynamic but this is a current workaround at the moment.

ajkiessl (changes requested):
Js linter is failing. You can run the linter locally with `yarn lint`.

ajkiessl (approved):
lgtm

PR #1413, opened 2025-04-11
Fix special request links for Microforms

memsdm05:
Current behavior requests raw.json for every link. Records with many locations will have lots of links, each requesting raw.json simultaneously. This topples smaller deployments like qa. This pr implements a cache that allows components to reuse a raw.json request locally.

### DIFF
--- a/app/javascript/availability/components/special_request_link.jsx
+++ b/app/javascript/availability/components/special_request_link.jsx
@@ -3,6 +3,8 @@ import { useEffect, useState } from 'react';
 import availability from '../index';
 import SpinnerLink from './spinner_link';
 
+const cache = {};
+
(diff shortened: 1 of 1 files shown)


app/javascript/availability/components/special_request_link.jsx
```diff
@@ -140,6 +157,7 @@ const SpecialRequestLink = ({ holding, locationText }) => {
     ) {
       text = 'Request Scan - Penn State Users';
     }
+    console.log(holding, hasData, cache, text);
```
ajkiessl, line 160: Logging statement need removed.  Other than that this lgtm

ajkiessl (approved):
LGTM.  Thanks!

PR #1417, opened 2025-04-28
Bootstrap 5 Upgrade

### LINKED ISSUE #1412: upgrade to bootstrap 5
this will resolve security vulnerabilities and make branding changes much easier

granta03:
Fixes #1412


granta03:
Current issues I found that still need to be resolved

- Two RSpec tests still need to be resolved. They involve the advanced search page. A Blacklight override may be required. 

- The dropdowns in Advanced Search don't work. I haven't been able to trace how exactly they are rendered though but some `data-*` attributes need to be replaced with the `data-bs-*` attribute that Bootstrap 5 requires.

- The text for the search fields in Advanced Search need to be aligned to the right. This can be done by replacing `text-md-right` to `text-md-end`. Like with dropdowns I haven't been able to trace how exactly these are rendered.

- The images for the books on the search results page are aligned to the left instead of the right

granta03:
Moved to #1445 . Preview branch

PR #1421, opened 2025-04-29
Handle nulls in bound_info.

ajkiessl:


ajkiessl:
Closing since we want this to fail if there's a null.  It indicates something wrong with the data

PR #1422, opened 2025-04-30
Allow Accessiblity Tools to See Cover Image

memsdm05:
Since adding alt text to cover images, the cover images found on listing pages where invisible to accessibility features. This PR adjusts how these specific images are presented so that they can still be recognized by various tools.


ajkiessl (approved):
lgtm

PR #1445, opened 2025-06-17
Upgrading Bootstrap 4 to Bootstrap 5

### LINKED ISSUE #1412: upgrade to bootstrap 5

granta03:
Fixes #1412 

Preview deployment: https://catalog-bootstrap-5.dev.k8s.libraries.psu.edu/


ajkiessl (commented):
A couple things I noticed:

-When hovering over a link, it does not underline or change color.  We want it to look like this: 
Screenshot 2025-06-17 at 4 30 37 PM

- The form field for the select dropdowns in the advanced search is constricted and the cursor is half cutoff.  Can we get this styled to look like the other fields on this form?
Screenshot 2025-06-17 at 4 21 43 PM
- The search button is rounded on all sides, but in production, it's only rounded on the right side.  
Screenshot 2025-06-17 at 4 26 48 PM

ajkiessl (approved):
Looks good nice work.  Ruth is out until the 23rd, so let's hold off on merging this until then.  We'll want her to review this before it goes out into prod.

app/javascript/blacklight_frontend_overrides/javascripts/modal.js
```diff
@@ -0,0 +1,15 @@
+// This override forces Blacklight to use Bootstrap 5's modal.
+// Future upgrades of Blacklight or Bootstrap may affect this.
+import Modal from 'bootstrap/js/dist/modal';
+
+Blacklight.modal.hide = function (el) {
+  Modal.getOrCreateInstance(
+    el || document.querySelector(Blacklight.modal.modalSelector),
+  ).hide();
+};
```
jlandiseigsti, line 9: This is a really good solution to the modal issue – nice work!

jlandiseigsti: The only thing to consider changing is to include this in our `blacklight_overrides` directory rather than creating a new `blacklight_frontend_overrides`. Other components in the 'blacklight_overrides' directory belong to blacklight_frontend...both `bookmark_toggle` and `checkbox_submit.` Then there would be one less `include` you'd need on `psulib_blacklight/index.js`

jlandiseigsti (changes requested):
We're really close to the finish line! (I've done a bootstrap upgrade and know they are a royal pain – thanks for your perserverance with this)

The Share and Bookmark buttons are now gray instead of teal...are we able to keep the teal?

The links in the footer no longer change color when hovered over, but I actually think we should keep that change. In production, the hovered links do not meet WCAG standards for color contrast. Let's keep it how you have it for now.

So my only notes are 1) moving where your override for the modal javascript lives, and to 2) making sure buttons such as Share match their original color.

granta03:
> The Share and Bookmark buttons are now gray instead of teal...are we able to keep the teal?
> 
@jlandiseigsti We can but Bootstrap 5 changed the btn-info class where it is teal to meet contrast guidelines. I talked with Alex about that which is why they are currently gray. This is how they would look if teal. 
Screenshot 2025-07-01 at 1 14 07 PM

jlandiseigsti:
Sorry, didn't know you had the button color discussion with Alex. I'm fine with the buttons being gray if that's better for contrast.

ajkiessl (approved):
LGTM.  We can have Ruth take a look next week.

PR #1485, opened 2025-11-11
downgrade blacklight_range_limit to fix pub year selector

Smullz622:
Publication year facet is broken in QA. This puts blacklight_range_limit back to last compatible version. Also has a very small mysql upgrade that resolved a build issue


jlandiseigsti (approved):
👍

PR #1489, opened 2025-11-14
Bugsnag ignore

ajkiessl:
I actually tested this locally to make sure it works this time.  Instead of ignoring bugsnag error via "path" use "railsAction".


jrpatterson (approved):
We will have to check our healthchecks across the apps to make sure we use the same terminology.

PR #1573, opened 2026-07-20
upgrade to rails 8.1

### LINKED ISSUE #1572: upgrade to Rails 8.1

Smullz622:
Fixes #1572 

Includes an upgrade to bot_challenge_page that was necessary for compatibility which involved significant change in the gem API


app/controllers/catalog_controller.rb
```diff
@@ -538,17 +538,6 @@ def trailing_punctuation?
...
-    def enforce_bot_challenge
-      # Challenge only if remote IP is not whitelisted
-      ip_whitelist = ENV.fetch('BOT_CHALLENGE_IP_WHITELIST', '')
-        .split(',')
-        .map { |ip| IPAddr.new(ip.strip) unless ip.strip.empty? }
-        .compact
-      return if ip_whitelist.any? { |ip| ip.include?(IPAddr.new(request.remote_ip)) }
-
-      BotChallengePage::BotChallengePageController.bot_challenge_enforce_filter(self, immediate: true)
-    end
```
ajkiessl, line 550: Was it intentional to remove the IP whitelist filter?  I think we might still need this for the bento connection, but I haven't touched that in a while so maybe something changed.

config/initializers/new_framework_defaults_8_1.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+# Be sure to restart your server when you modify this file.
+#
+# This file eases your Rails 8.1 framework defaults upgrade.
+#
+# Uncomment each configuration one by one to switch to the new default.
+# Once your application is ready to run with all new defaults, you can remove
+# this file and set the `config.load_defaults` to `8.1`.
```
ajkiessl: We should try uncommenting the configurations here and load the rails 8.1 defaults in the application.rb to see if these work.

ajkiessl (changes requested):

config/initializers/new_framework_defaults_8_1.rb
```diff
@@ -0,0 +1,76 @@
+ # frozen_string_literal: true
```
ajkiessl: This file can actually be removed.  Looks like if the `load_defaults` in the application.rb is set to 8.1, then these configurations automatically get loaded.

spec/controllers/catalog_controller_spec.rb
```diff
@@ -27,42 +27,6 @@
...
-        it 'calls bot_challenge_enforce_filter' do
-          request.remote_addr = non_whitelisted_ip
-
-          get :index
-          expect(BotChallengePage::BotChallengePageController)
-            .to have_received(:bot_challenge_enforce_filter)
-            .with(instance_of(described_class), immediate: true)
-        end
-      end
-    end
```
ajkiessl, line 65: I think we should still have these tests to make sure the behavior is correct.  This is how I handled testing the IP whitelist with the new syntax in Explore:

https://github.com/psu-libraries/etda_explore/blob/main/spec/controller/files_controller_spec.rb#L44-L67

ajkiessl (changes requested):

ajkiessl (approved):
LGTM!

